PR #6, opened 2018-12-21
Add SocketIO client

cjh1:


stempy/image.cpp
```diff
@@ -0,0 +1,67 @@
+#include "image.h"
+#include "mask.h"
+
+#include <memory>
+
+using namespace std;
```
cryos: We should try to get in the habit of just bringing in the symbols we use instead of the whole std namespace.

stempy/reader.h
```diff
@@ -16,11 +16,21 @@ struct EofException : public std::exception
...
+
+  Header() = default;
+  Header(const Header& header) = default;
+  Header(Header&& header) noexcept = default;
 };
 
 struct Block {
   Header header;
-  std::unique_ptr<uint16_t[]> data;
+  std::shared_ptr<uint16_t> data;
```
cryos: Why shared_ptr over unique_ptr?

stempy/image.cpp
cjh1: Sure, I can do that.

stempy/reader.h
cjh1: Memory management gets a little hair when using the Block from the python side, so for now I switch to shared_ptr. When the use-case for the API get a little more define, we may be able to switch back ...

PR #9, opened 2019-04-02
Add support for new file format

cjh1:


cjh1:
@cryos This allows use to read the new file format. I am getting very high values for the first image number in the header, not sure why.

PR #11, opened 2019-04-08
Add parallel STEM image generation using VTKm

psavery:


psavery:
It would be nice if we could find some ways to be more concise/elegant with the [parallel sum](https://github.com/OpenChemistry/stempy/pull/11/files#diff-f9181c094fe925b29551c9fbb04e34e9R46) function. I feel that it is a bit verbose. But maybe there isn't a way to make it shorter.

One problem that I have not figured out how to cope with: I know that the sums such as stemValues.dark are `uint64_t` so that they can store larger values. But the `ReduceByKey` function in vtkm requires that the input type and output type be the same. Since the input type is `uint16_t`, I have to make the sums `uint16_t` as well, which is probably going to result in overflow, right?

cjh1:
@robertmaynard Is there a way for the output type for the `ReduceByKey` to be different to the input type? See @psavery comment above.

stempy/image.cpp
```diff
@@ -36,6 +42,55 @@ STEMValues calculateSTEMValues(uint16_t data[], int offset,
...
+  auto keysBright = vtkm::cont::make_ArrayHandle(brightFieldMask,
+                                                 numberOfPixels);
+  auto keysDark = vtkm::cont::make_ArrayHandle(darkFieldMask, numberOfPixels);
+  auto input = vtkm::cont::make_ArrayHandle(&data[offset], numberOfPixels);
+
+  vtkm::cont::ArrayHandle<uint16_t> uniqueKeys;
+  vtkm::cont::ArrayHandle<uint16_t> sums;
+
+  // Portals so we may view the values
+  auto uniqueKeysPortal = uniqueKeys.GetPortalConstControl();
```
robertmaynard: You should get the portals after the algorithm has executed, not before.
You should presume that all data parallel algorithm are executing in an async manner and that the algorithm will execute intil another algorithm execution, or `ArrayHandle::GetPortalConstControl` is called.

robertmaynard:
I would use an ArrayHandleCast on the input to make it act like `uint64_t` instead of an `uint16_t`.

stempy/image.cpp
psavery: Oh okay. Will do.

robertmaynard:
It looks like you are computing the bright/dark for a single image in a single invocation. For that kind of work I would use the following:

```cpp

struct MaskAndAdd
{
  VTKM_EXEC_CONT
  uint64_t operator()(const vtkm::Pair<uint16_t,uint16_t>& a, const vtkm::Pair<uint16_t,uint16_t>& b) const
  {
    return (a.first & a.second) + (b.first & b.second);
  }

  VTKM_EXEC_CONT
  uint64_t operator()(uint64_t a, uint64_t b) const
  {
    return a + b;
  }

  VTKM_EXEC_CONT
  uint64_t operator()(const vtkm::Pair<uint16_t,uint16_t>& a, uint64_t b) const
  {
    return (a.first & a.second) + b;
  }

  VTKM_EXEC_CONT
  uint64_t operator()(uint64_t a, const vtkm::Pair<uint16_t,uint16_t>& b) const
  {
    return a + (b.first & b.second);
  }
};


  vtkm::cont::ArrayHandle<uint16_t> input = vtkm::cont::make_ArrayHandle(&data[offset], numberOfPixels);
  vtkm::cont::ArrayHandle<uint64_t> scratch;

  auto inputAndBright = vtkm::cont::ArrayHandleZip(input, keysBright);
  auto inputAndDark = vtkm::cont::ArrayHandleZip(input, keysDark);

  uint64_t brightSum = vtkm::cont::Algorithm::Reduce(inputAndBright, scratch, MaskAndAdd{} );
  uint64_t darkSum = vtkm::cont::Algorithm::Reduce(inputAndDark, scratch, MaskAndAdd{} );


```

From that point we can either look at fusing the two Reduce calls into a single call, or we can look at executing across multiple images in the same call.

psavery:
@robertmaynard Sounds good. I'll try it out!

psavery:
I re-built VTK-m and stempy in Release mode, which made a big difference for a benchmark.

I used the `time` module from python to time [this function](https://github.com/OpenChemistry/stempy/blob/54bc487e213c76a860f959f88f6f2347b92c759e/examples/stem_image.py#L26) once for serial mode and once for parallel mode on a 20 core computer.

Serial mode ran in 4.20 seconds, and parallel mode ran in 1.49 seconds.

robertmaynard:
Debug builds for VTK-m will generally have poor performance as we rely on the optimization pass/inlinling to remove levels of abstraction.
But there are some user controls that will also help improve runtime perf:
1. Use DebWithRelInfo as your build type as that will improve the optimization level
2. Set the build option VTKm_NO_ASSERT to ON.

psavery:
I combined the two masks into one in the most recent commit. I got a performance boost, 0.67 seconds (averaged over 10 times) with the combined masks, and 0.95 seconds without.

I'm wondering if it would be better to use `ArrayHandleCompositeVector` though...

cjh1:
@psavery Lets try that approach as well.

psavery:
The last commit uses `ArrayHandleCompositeVector`. I think it's more readable, but I still need to benchmark it...

cjh1:
@psavery I agree, if it performs OK, lets go with that.

psavery:
@robertmaynard When I compile [this file](https://github.com/OpenChemistry/stempy/blob/522a1f15fcb4c684e64f08f4b308e51c431a2620/stempy/image.cpp) with gcc (using vtkm with OpenMP), I don't get any errors, and it seems to run fine. But when I compile it with cuda (using vtkm with CUDA), I get the attached [error](https://github.com/OpenChemistry/stempy/files/3064852/error.txt). It mentions that when it is instantiating my [reduce function](https://github.com/OpenChemistry/stempy/pull/11/files#diff-f9181c094fe925b29551c9fbb04e34e9R107), it fails to convert `vtkm::Vec<uint16_t, 3>` (the input type) to `vtkm::Pair<uint64_t, uint64_t>` (the output type). 

I don't understand why it is trying to convert the input type to the output type, though. I will keep looking at it. Do you know what the issue might be, though?

robertmaynard:
@psavery This looks to be a deficiency in the cuda implementation where we don't handle vtkm::Pair<> as a reduction value type properly.

I believe that if you switch from a result type of `vtkm::Pair<>` to a `vtkm::Vec<>` you should be able to compile with cuda ( and necessary operator changes ). While you verify that I can create an issue on VTK-m for adding pair support to cuda reduce.

psavery:
Unfortunately, it looks like I get the same kind of error with `vtkm::Vec<>`. It is
[attached](https://github.com/OpenChemistry/stempy/files/3065000/error2.txt).

I'll make a commit that shows my changes to the `CMakeLists.txt` so that we can see that too.

psavery:
Gonna force push to remove the last commit. We will put the CUDA stuff in another PR.

psavery:
Thanks @cjh1!

robertmaynard:
@psavery This is most definitely a deficiency inside the VTK-m CUDA reduction implementation. I think I know how we can solve the issue, but it will require modifications to VTK-m itself.

psavery:
@robertmaynard Okay. We can put this on hold until that gets fixed. Let us know when.

robertmaynard:
@psavery Can you see if applying these patches to VTK-m fixes the issue for you? ( https://gitlab.kitware.com/vtk/vtk-m/merge_requests/1638 ).

You will need to modify the binary operator to also be able to transform a single input value to a single output value too, as documented [in the changelog](https://gitlab.kitware.com/vtk/vtk-m/blob/89ec4aae2fa2188df7eebd5189f02c6beca9b20b/docs/changelog/reduction-support-differing-input-output-types.md).

psavery:
I am indeed able to compile it with those changes!

I'm gonna give it a test run as well.

PR #12, opened 2019-04-09
Add serial electron counting

cjh1:


stempy/electron.cpp
```diff
@@ -0,0 +1,268 @@
...
+  double variance = 0;
+
+  for (int i = 0; i < values.size(); i++) {
+    variance += pow(values[i], 2.0);
+  }
+
+  return (variance / (values.size() - 1)) - pow(mean, 2.0);
+}
+
+std::pair<int16_t, int16_t> findMinMax(std::vector<int16_t>& values)
```
psavery: I think we could use STL for this function. Something like
```
  auto minmax = std::minmax_element(values.begin(), values.end());
  return std::make_pair(*minmax.first, *minmax.second);
```

stempy/electron.cpp
```diff
@@ -0,0 +1,268 @@
...
+    }
+    if (values[i] > max) {
+      max = values[i];
+    }
+  }
+
+  return std::make_pair(min, max);
+}
+
+uint64_t findMax(std::vector<int16_t>& values)
```
psavery: Similarly for this one:
```
return *std::max_element(values.begin(), values.end());
```

cjh1: Not used anyway, will remove

PR #14, opened 2019-04-10
Compile with CUDA if CUDA was enabled in vtkm

psavery:
This is based upon the contour_tree_augmented example
in vtkm.


psavery:
This has two compilation errors: [error1.txt](https://github.com/OpenChemistry/stempy/files/3065122/error1.txt) and [error2.txt](https://github.com/OpenChemistry/stempy/files/3065124/error2.txt)

We may remove the function in error2.txt that produces the compilation error, so that one may not need to be fixed. But we do need to fix the compilation error in error1.txt.

psavery:
@cjh1 @robertmaynard This is the new PR for using CUDA.

psavery:
With the new commit and the vtk-m MR mentioned in it, error1 no longer occurs.

If the process function [here](https://github.com/OpenChemistry/stempy/blob/7e0ce303caf3cdb72db8915251f835a8187b4f3b/stempy/reader.h#L44) is also removed, along with its definitions and uses, the program compiles with CUDA.

psavery:
@cjh1 I merged your PR to my commit here. I'm gonna delete my fork of stempy.

psavery:
@cjh1 @robertmaynard

Following the recent changes, I benchmarked all three types of runs on ulex. Each benchmark is an average over 10 runs. There are 20 processes for OpenMP. Here are the results:
```
Serial: 3.90 seconds
OpenMP: 0.67 seconds
CUDA: 4.71 seconds
```

I used md5sum to ensure they all produced the same output, and they did.

So, it looks like OpenMP is still very fast. CUDA is much faster than it was, but it is still a little slower than serial.

robertmaynard:
For CUDA we have converted the problem being computationally bound to be IO bound. In our case we are stalling on the CPU before the reduce calls as we wait for the data to transfer over PCI-E. This occurs as the initiation of transfer happens immediately before execution of the algorithm.

What we would want is to have the ability to transfer the images to the GPU while we load them from disk  in a background worker. That would allow us to hide the transfer time and have the algorithm be computationally bound on CUDA.

cjh1:
The whole problem is pretty IO bound ( all the times are excluding the load time, which is a significant part of the wall clock time ). One of the changes I would like to make soon now that we have a few algorithms implemented is to use iterators to stream the blocks into these functions so that we can overlap the IO with the computation. One this is done I think it will be a good time to revisit this and try a few more optimizations.

CMakeLists.txt
```diff
@@ -41,8 +41,15 @@ if(VTKm_FOUND)
   list(APPEND _libs vtkm_cont)
 endif()
 
-add_library(stem
-  ${_src})
+if(TARGET vtkm::cuda)
+  # Compile with CUDA
+  vtkm_compile_as_cuda(cudaSource ${_src})
+  add_library(stem ${cudaSource})
+  target_compile_definitions(stem PRIVATE
```
robertmaynard: Setting the default device adapter isn't needed anymore, and is a legacy define we are currently removing.

CMakeLists.txt
```diff
@@ -53,9 +60,15 @@ endif()
 target_link_libraries(stem
   PRIVATE ${_libs})
 
-add_executable(stemcli
-  stempy/main.cpp
-)
+if(TARGET vtkm::cuda)
```
robertmaynard: I don't think that the main needs to be compiled with cuda. It is fine to have this as just cxx

PR #15, opened 2019-04-10
Make turning on VTK-m support explicit

cjh1:


CMakeLists.txt
```diff
@@ -36,8 +36,9 @@ endif()
 
 set(_libs Threads::Threads)
 
-find_package(VTKm COMPONENTS OpenMP)
-if(VTKm_FOUND)
+option(stempy_ENABLE_VTKm "Build with VTK-m" OFF)
+if (${stempy_ENABLE_VTKm})
```
psavery: I think this can just be:
```
if (stempy_ENABLE_VTKm)
```

without the curly braces. See [here](https://gitlab.kitware.com/vtk/vtk-m/blob/master/CMakeLists.txt#L156), for instance.

cjh1: Yep, good catch. Done

PR #17, opened 2019-04-11
Add parallel method to get maximal pixels

psavery:
It still needs some work and improvement in a few
areas, but it seems to function.

It is based upon the game_of_life example in vtkm.

My timings for 10-run averages on my 6-core computer:

Serial: 6.63 seconds
OpenMP: 5.56 seconds


psavery:
To do: I think [this](https://github.com/OpenChemistry/stempy/pull/17/files#diff-5afbef7ae6ecfa409c949d37ebfd1e93R68) can be a smaller data type since we are treating it as a bool. We should be able to `vtkm::UInt8` instead.

stempy/electron.cpp
```diff
@@ -229,9 +358,14 @@ std::vector<std::vector<std::pair<int, int>>> electronCount(
         }
```
cjh1: We should do the thresholding in the same pass, otherwise we are still looping through each pixel twice?

robertmaynard:
The game of life showcases the filter infrastructure so that we can demo the filter API, for this problem I would not use that as you have no need for dynamic type support. Instead I would use the following:



```cpp
#include "config.h"

 #ifdef VTKm
#include <vtkm/cont/ArrayHandle.h>
#include <vtkm/cont/DataSetBuilderUniform.h>
#include <vtkm/filter/FilterDataSet.h>
#include <vtkm/worklet/DispatcherPointNeighborhood.h>
#include <vtkm/worklet/WorkletPointNeighborhood.h>
#endif

 #ifdef VTKm
namespace {


 struct IsMaximalPixel : public vtkm::worklet::WorkletPointNeighborhood
{
  using CountingHandle = vtkm::cont::ArrayHandleCounting<vtkm::Id>;

   using ControlSignature = void(CellSetIn,
                                FieldInNeighborhood neighborhood,
                                FieldOut isMaximal);

   using ExecutionSignature = void(_2, _3);

   template <typename NeighIn>
  VTKM_EXEC void operator()(const NeighIn& neighborhood,
                            bool& isMaximal) const
  {
    isMaximal = 0;

    auto current = neighborhood.Get(0, 0, 0);
    for (int j = -1; j < 2; ++j) {
      for (int i = -1; i < 2; ++i) {
        if (i == 0 && j == 0)
          continue;
        if (current <= neighborhood.Get(i, j, 0))
          return;
      }
    }

     isMaximal = 1;
  }
};

 std::vector<std::pair<int, int>> maximalPointsParallel(
  const std::vector<uint16_t>& frame, int rows, int columns)
{
  // Build the data set
  vtkm::cont::CellSetStructured<2> cellset(rows,columns,1);  

  auto frameHandle = vtkm::cont::make_ArrayHandle(frame);
  vtkm::cont::ArrayHandle<bool> maximalPixels;

  vtkm::worklet::Invoker invoke;
  invoke(IsMaximalPixel{}, cellset, frameHandle, maximalPixels)

   // Convert to std::vector<std::pair<int, int>>
  auto maximalPixelsPortal = maximalPixels.GetPortalConstControl();
  std::vector<std::pair<int, int>> outputVec;
  outputVec.reserve(maximalPixelsPortal.GetNumberOfValues());
  for (vtkm::Id i = 0; i < maximalPixelsPortal.GetNumberOfValues(); ++i) {
    if (maximalPixelsPortal.Get(i) == 1) {
      auto row = i / columns;
      auto column = i % columns;
      outputVec.push_back(std::make_pair(row, column));
    }
  }

   // Done
  return outputVec;
}
} // end namespace
#endif
```

Note that I inverted the `for loops` as i should increase faster than j.

psavery:
@robertmaynard Thank you for the feedback. This does simplify things a lot. Will benchmark again soon...

psavery:
I found an issue with the old timings in the first commit message, so they are not necessarily comparable.

My new timings with the latest commit:
```
Serial: 6.53 seconds
OpenMP: 4.52 seconds
```

They are averaged over 5 runs on a computer with 6 physical cores with hyperthreading (12 processes).

I'm gonna look into two ways to parallelize the thresholding operations now.

psavery:
The latest commit resulted in a performance boost. It now runs in ~2.79 seconds in parallel.

I'm also going to try modifying the frame directly as a separate parallel operation, and see how that goes. I also have a few other small ideas for performance boosts.

psavery:
The most recent commit, performing the background subtraction and thresholding together as a separate, parallel step is even faster. It was ~1.13 seconds for me.

psavery:
Unfortunately, I can't compile the file with CUDA. It looks like there may be some issues with eigen 3.3.4-4 when compiling with CUDA. Maybe I can separate our parallel functions into another file.

psavery:
CUDA now compiles `electron.cpp` as of the latest commit.

Here are some timings that I performed as of the latest commit:
```
Serial time: 6.53 seconds
OpenMP time: 1.15 seconds
CUDA time: 4.86 seconds
```

Additionally, there is a separate commit I made and put in another branch where the CUDA performance is enhanced, similar to how we enhanced the CUDA performance of `image.cpp`. That commit is [here](https://github.com/OpenChemistry/stempy/commit/bc050915caffad3e446c4d97458d13ea42483aa3). The timings for that commit:
```
Serial time: 6.57 seconds
OpenMP time: 1.43 seconds
CUDA time: 2.71 seconds
```

Note that CUDA performance increased, but the OpenMP performance decreased. I have not merged that commit with this branch yet because I am not sure if it is worth the OpenMP performance decrease. I'll wait for feedback before doing so.

All timings were averages of 5 runs performed on my 6 core computer with hyperthreading (12 processes). My video card is a GeForce MX130.

psavery:
@cjh1 I updated the `electron_count` API to use the iterators. Note that the threshold calculation API [still uses the vector of blocks](https://github.com/OpenChemistry/stempy/pull/17/files#diff-95d11f46d7139239956042d4e30ff1dfR42) because I figured that the caller can put a few blocks together of their choosing and then pass it in as a vector of blocks (in our sample, we have just one block, so it doesn't matter too much).

One thing we may want to do in the future is calculate the maximal pixels in separate threads using a ThreadPool while the main thread reads in the data, similar to what we did for the stem image. It doesn't matter for our current example, though since there is only one block.

stempy/reader.h
```diff
@@ -45,6 +45,9 @@ class StreamReader {
   void process(int streamId, int concurrency=-1, int width=160, int height=160,
       const std::string& url="http://127.0.0.1:5000");
 
+  // Rewind to the start of the first file
+  void rewind();
```
cjh1: Lets call the `reset`

python/io.cpp
```diff
@@ -39,6 +39,7 @@ PYBIND11_MODULE(_io, m)
     .def(py::init<const std::string&, uint8_t>())
     .def(py::init<const std::vector<std::string>&, uint8_t>())
     .def("read", (Block(StreamReader::*)()) & StreamReader::read)
+    .def("rewind", &StreamReader::rewind)
```
cjh1: Lets call this reset.

PR #18, opened 2019-04-14
Added a block iterator for createSTEMImage

psavery:
This block iterator is an input iterator that iterates
over blocks one at a time from the list of files it has been
given. This allows the program to avoid loading all of the
data in at once, and instead, each block is loaded when it is
needed.

Some work probably needs to be done. I'm not convinced that
my design of the BlockIterator class is the best.

I haven't done timings yet, either. But one good thing is that I can
now run the stem_example.py on my computer for our sample data, 
because all 16 gigabytes of data no longer need to be loaded at once.


stempy/blockiterator.cpp
```diff
@@ -0,0 +1,66 @@
...
+namespace stempy {
+
+BlockIterator::BlockIterator(const vector<string>& fileList)
+{
+  m_streams.reserve(fileList.size());
+  for (const auto& file: fileList)
+    m_streams.push_back(StreamReader(file));
+
+  if (m_streams.empty())
+    cerr << "Warning: no files provided for BlockIterator" << endl;
```
cjh1: For this sort of error we should throw an exception.

stempy/blockiterator.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+  ++m_numBlocksRead;
+
+  return *this;
+}
+
+bool BlockIterator::atEnd() const
+{
+  return m_curStreamIndex >= m_streams.size();
+}
```
cjh1: We want this to be a standard [input iterator](http://www.cplusplus.com/reference/iterator/InputIterator/), as far as I am aware `atEnd` is not part of that. We should also as the appropriate tag, as shown in the example in that link.

stempy/blockiterator.cpp
```diff
@@ -0,0 +1,66 @@
...
+using std::cout;
+using std::cerr;
+using std::endl;
+
+using std::string;
+using std::vector;
+
+namespace stempy {
+
+BlockIterator::BlockIterator(const vector<string>& fileList)
```
cjh1: I was morning thinking that we would be able to iterator over the reader, something like:
```c++
StreamReader::iterator iter = reader.begin()

```

We can overload the StreamReader constructor to take a list of files.

stempy/image.cpp
```diff
@@ -113,13 +114,20 @@ STEMValues calculateSTEMValuesParallel(
...
 
-STEMImage createSTEMImage(std::vector<Block>& blocks, int rows, int columns,  int innerRadius, int outerRadius)
+template <typename InputIt>
+STEMImage createSTEMImage(InputIt first, InputIt last, int rows, int columns,
+                          int innerRadius, int outerRadius)
 {
   STEMImage image(rows, columns);
 
+  if (first == last) {
+    cerr << "Error in " << __FUNCTION__ << ": no blocks to read!\n";
```
cjh1: This should be an exception

stempy/reader.h
```diff
@@ -39,15 +39,78 @@ class StreamReader {
...
   StreamReader(const std::string &path, uint8_t version=1);
+  StreamReader(const std::vector<std::string>& files, uint8_t version = 1);
 
-  Block read();
+  Block& read();
   void process(int streamId, int concurrency=-1, int width=160, int height=160,
       const std::string& url="http://127.0.0.1:5000");
 
+  // Whether or not we are at the end of all of the files
+  bool atEnd() const { return m_curFileIndex >= m_files.size(); }
```
cjh1: Can this be private?

examples/benchmark_stem_image.py
```diff
@@ -0,0 +1,23 @@
...
+import time
+
+# Beware of disk caching when using this benchmark
+num_runs = 10
+times = []
+for i in range(num_runs):
+    start = time.time()
+
+    files = []
+    for f in glob.glob('/data/4dstem/smallScanningDiffraction/data*.dat'):
```
cjh1: Can we using [click](http://click.palletsprojects.com/en/7.x/) to allow the path to be passed in?

stempy/image.cpp
```diff
@@ -131,28 +179,43 @@ STEMImage createSTEMImage(std::vector<Block>& blocks, int rows, int columns,  in
...
-    for(int i=0; i<block.header.imagesInBlock; i++) {
+  // Run the calculations in a thread pool while the data is read from
+  // the disk in the main thread.
+  int numThreads = 4;
+  ThreadPool pool(numThreads);
+
+  // Populate the worker pool
+  vector<future<void>> futures;
+  for (; first != last; ++first) {
+    Block b = std::move(*first);
```
cjh1: I don't think this move is needed, we are already doing a move @ 196.

stempy/image.cpp
```diff
@@ -131,28 +179,43 @@ STEMImage createSTEMImage(std::vector<Block>& blocks, int rows, int columns,  in
...
 #ifdef VTKm
-      // Use view to the array already transfered
-      auto view = vtkm::cont::make_ArrayHandleView(
-        dataHandle, i * numberOfPixels, numberOfPixels);
-      auto stemValues = calculateSTEMValuesParallel(view, bright, dark);
+    // Instead of calling _runCalculateSTEMValues directly, we use a
+    // lambda so that we can explicity delete the block. Otherwise,
+    // the block will not be deleted until the threads are destroyed.
+    futures.emplace_back(pool.enqueue(
+      [b{ std::move(b) }, numberOfPixels, &bright, &dark, &image]() mutable {
```
cjh1: ` b{ std::move(*first)}`?

stempy/image.cpp
```diff
@@ -186,5 +249,9 @@ Image<double> calculateAverage(std::vector<Block> &blocks)
   return image;
 }
 
-
+// Instantiate the ones that can be used
+template STEMImage createSTEMImage(StreamReader::iterator first,
+                                   StreamReader::iterator last, int rows,
+                                   int columns, int innerRadius,
+                                   int outerRadius);
```
cjh1: Not sure I understand what is going on here?

stempy/reader.cpp
```diff
@@ -103,15 +139,26 @@ Header StreamReader::readHeaderVersion2() {
   return header;
 }
 
-Block StreamReader::read() {
+Block& StreamReader::read()
```
cjh1: Why the change in return type? The ownership semantic aren't clear here. Before the caller took ownership, here I'm not so sure? It seems it might always be a reference to `m_curBlock`, which would then change as the iteration continues, which is not what we want?

stempy/reader.h
```diff
@@ -39,15 +39,78 @@ class StreamReader {
...
+  iterator end() { return iterator(nullptr); }
+
+  class iterator
+  {
+  public:
+    using self_type = iterator;
+    using value_type = Block;
+    using reference = Block&;
+    using pointer = Block*;
+    using iterator_category = std::input_iterator_tag;
```
cjh1: Looks like you aren't using this?

psavery: Not personally, I was just trying to follow all of the requirements for making an input iterator, and I think this is one of them. If we pass it into an STL algorithm, it might use it. I could potentially use it by checking it in the `createSTEMImage` function to make sure it is an input_iterator_tag.

stempy/reader.cpp
psavery: This was done for the `StreamReader::iterator`.

For input iterators, `++it` is supposed to increment the iterator and then return the iterator. In our case, the increment is done by calling `StreamReader::read()`. If we don't keep a copy of the current block inside the `StreamReader` class, then the block will be destroyed immediately after creation when `++it` is called. We also need to store a copy of the current block inside the `StreamReader` class so that `*it` will return the current block. 

So now, the `StreamReader` class has ownership of the current block, and it returns references when possible to avoid making too many copies. However, multiple copies of a `block` can be made (multiple ownership), and the pointer will be deleted when they all get destroyed (thanks to `shared_ptr`). If we want to take ownership of the block from `StreamReader`, we can now use `std::move` on either the return of `read()` or the return of `*it`.

stempy/image.cpp
psavery: Since we want to use iterators for `createSTEMImage`, it means that we need to make it a template function. Unfortunately, because it is a template function, it complicates the implementation. We have two choices:

1. Expose all of the implementation of `createSTEMImage` in the header file. This would currently also mean exposing some of the VTKm parts as well (although we could also make another function and move them to it), and including `ThreadPool` and many other files in the header file. A lot of implementation would be included and exposed in the `image.h` header file, and I wasn't sure if that's what we wanted.

2. Explicitly instantiate the template for all of the types we are allowing. This is what is happening above, and is the option I chose, just because it is simpler for now. It also effectively whitelists the function so that only the types we instantiate can be used. Right now, the template type `InputIt` can only be `StreamReader::iterator`, and possibly other types if we add them to the list. If we want to allow any user-defined type to be used with `createSTEMImage`, then we'll need to go with option number 1.

stempy/image.cpp
psavery: Okay.

stempy/image.cpp
psavery: Okay.

examples/benchmark_stem_image.py
psavery: Sure!

stempy/image.cpp
cjh1: Ok, lets add `std::vector<Block>::iterator`?

stempy/reader.cpp
cjh1: Just so I understand. In the case that the using does the following:

```
Block& blockA = reader.read();
Block& blockB = reader.read();
```

Do `blockA` and 'blockB' now point to the same block?

stempy/reader.h
cjh1: Ok, thanks, I thought the tagging was done differently.

stempy/reader.cpp
psavery: Yeah, I think they do.

I was thinking about it some last night. If you think it'd be better to return `Block` from `read()`, we can make it so that `read()` doesn't set `m_curBlock`. But we can add another function or something called `increment` or something, that calls `read()` and sets `m_curBlock`. Then the `StreamReader::iterator` can use that function instead of `read()`. Just an idea.

stempy/image.cpp
```diff
@@ -131,28 +179,43 @@ STEMImage createSTEMImage(std::vector<Block>& blocks, int rows, int columns,  in
...
+    Block b;
 #ifdef VTKm
-      // Use view to the array already transfered
-      auto view = vtkm::cont::make_ArrayHandleView(
-        dataHandle, i * numberOfPixels, numberOfPixels);
-      auto stemValues = calculateSTEMValuesParallel(view, bright, dark);
+    // Instead of calling _runCalculateSTEMValues directly, we use a
+    // lambda so that we can explicity delete the block. Otherwise,
+    // the block will not be deleted until the threads are destroyed.
+    futures.emplace_back(pool.enqueue([b{ std::move(*first) }, numberOfPixels,
```
psavery: This doesn't compile on CUDA by the way. It complains about auto type deduction for `b{ std::move(*first) }`. I'm gonna look into fixing it.

stempy/image.cpp
psavery: Okay!

examples/benchmark_stem_image.py
```diff
@@ -0,0 +1,36 @@
...
+import time
+
+# Beware of disk caching when using this benchmark
+@click.command()
+@click.argument('path', type=click.Path(exists=True, file_okay=False))
+@click.option('-n', '--num-runs', default=10, help='Number of runs to perform')
+def run_benchmarks(path, num_runs):
+    """Run benchmarks on a directory containing data*.dat files"""
+    files = []
+    for f in glob.glob(path + '/data*.dat'):
```
cjh1: I was thinking that we should be able to pass in the whole glob expression, not sure we can rely on the file name starting with data.

examples/benchmark_stem_image.py
```diff
@@ -0,0 +1,36 @@
...
+        end = time.time()
+        times.append(end - start)
+        print('Run ' + str(len(times)) + ': {:0.2f} seconds'.format(times[-1]))
+
+    print('Number of runs was:', num_runs)
+    print('Average time: {:0.2f} seconds'.format(sum(times) / num_runs))
+
+
+if __name__ == '__main__':
+    run_benchmarks()
```
cjh1: This is looking nice

stempy/reader.cpp
cjh1: Too me that is a little confusing, lets chat about this.

cryos: This sounds like it is going to be a possible source of subtle bugs, and we should make the iterator store the current block, not overload the read method. Just catching up on some of what has been happening, but this doesn't look intuitive in its current state.

psavery: Yeah, I'm fixing the `read()` to return a `Block`, and moving the current block into the iterator now. We started looking at `std::istream_iterator` as a reference, as it seems fairly similar to what we are doing.

psavery:
For the record, and in case we want to look at this for future reference, here are some timings I obtained on a 10 core computer with hyperthreading (20 processors) on a Quadro K4200 GPU. These were obtained a few commits back, on commit 480192dc46a9b9cf72bcbeff99a2741bfd23ddb7. Each run is an average of 10 runs.

****Master Branch****

***SSD***
Serial: 38.00 seconds
OpenMP: 35.09 seconds
CUDA:   39.38 seconds

***RAMDisk***
Serial: 11.87 seconds
OpenMP:  9.01 seconds
CUDA:   13.37 seconds

****Current Commit****

***SSD***
Serial: 33.64 seconds
OpenMP: 33.66 seconds
CUDA:   33.81 seconds

***RAMDisk***
Serial: 4.43 seconds
OpenMP: 4.01 seconds
CUDA:   5.78 seconds

cryos:
It would be good to get timings on a better CUDA capable system, in general most of these timings look to still be failing to scale well likely due to IO. As we look to sparse data pipelines in addition to raw data it will be interesting to see how this evolves too. The fact that all three are working is also a very useful line in the sand.

PR #20, opened 2019-04-17
Return electron count position as a 1D array index

cjh1:
This is a simpler representation rather than have to pass around
std::pair.

### DIFF
--- a/examples/electron.ipynb
+++ b/examples/electron.ipynb
@@ -51,8 +51,8 @@
     "# Just sum the events for now\n",
     "for frame in frame_events:\n",
     "    for pos in frame:\n",
-    "        row = pos // 576\n",
-    "        column = pos % 576\n",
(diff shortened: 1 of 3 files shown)


examples/electron.ipynb
```diff
@@ -51,7 +51,9 @@
     "# Just sum the events for now\n",
     "for frame in frame_events:\n",
     "    for pos in frame:\n",
-    "        image[pos[0]][pos[1]] += 1\n"
+    "        row = pos // 576\n",
```
psavery: Can we figure out `576` here programmatically?

cjh1: Not real, its just the physical size of the detector.

psavery: Okay. Yeah, I kind of wish that we could. But it's okay.

psavery:
Looks great to me!

PR #21, opened 2019-04-17
Image number indexes start from one

cjh1:


psavery:
@cjh1 Looks like it is this PR that is causing the seg fault. If I revert master back before this one and compile, it runs fine.

psavery:
Probably because of the `header.imageNumbers.push_back(firstImageNumber + i)` down below... maybe I can fix that.

psavery:
I think I found the cause of the crash. It occurs [here](https://github.com/OpenChemistry/stempy/blob/f7363b38ba1abca180482d7e6b2c1f0b621e04a7/stempy/electron.cpp#L233), because the image numbers no longer start with 0.

cjh1:
@psavery Great debugging, thanks, I will look at pushing a fix.

PR #23, opened 2019-04-19
Ensure image values are set to zero

cjh1:
This is important with a partial image.

### DIFF
--- a/stempy/image.cpp
+++ b/stempy/image.cpp
@@ -30,6 +30,9 @@ STEMImage::STEMImage(uint32_t width, uint32_t height)
 {
   this->bright = Image<uint64_t>(width, height);
   this->dark = Image<uint64_t>(width, height);
+  // Init values to zero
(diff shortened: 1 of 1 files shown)


stempy/image.cpp
```diff
@@ -30,6 +30,9 @@ STEMImage::STEMImage(uint32_t width, uint32_t height)
 {
   this->bright = Image<uint64_t>(width, height);
   this->dark = Image<uint64_t>(width, height);
+  // Init values to zero
```
psavery: Yeah, I guess we could either do this here, or in the `Image` constructor right above. If you think we might need it for other use cases of `Image`, maybe we should put it there. Otherwise, here is good.

cjh1: yes, I thought about that, but in other cases we may not need the over head of initialization if we know we are going to assign a value to every pixel.

PR #26, opened 2019-04-26
Use StreamReader::iterator for calculateAverage

psavery:

### DIFF
--- a/stempy/image.cpp
+++ b/stempy/image.cpp
@@ -228,16 +228,18 @@ STEMImage createSTEMImage(InputIt first, InputIt last, int rows, int columns,
   return image;
 }
 
-Image<double> calculateAverage(std::vector<Block> &blocks)
+template <typename InputIt>
+Image<double> calculateAverage(InputIt first, InputIt last)
 {
-  auto detectorImageRows = blocks[0].header.rows;
-  auto detectorImageColumns = blocks[0].header.columns;
+  auto detectorImageRows = first->header.rows;
+  auto detectorImageColumns = first->header.columns;
   auto numberOfPixels = detectorImageRows*detectorImageColumns;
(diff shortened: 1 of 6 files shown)


stempy/image.cpp
```diff
@@ -228,16 +228,18 @@ STEMImage createSTEMImage(InputIt first, InputIt last, int rows, int columns,
...
+  auto detectorImageRows = first->header.rows;
+  auto detectorImageColumns = first->header.columns;
   auto numberOfPixels = detectorImageRows*detectorImageColumns;
   Image<double> image(detectorImageRows, detectorImageColumns);
 
   std::fill(image.data.get(), image.data.get() + numberOfPixels, 0.0);
   uint64_t numberOfImages = 0;
-  for(const Block &block: blocks) {
+  for (; first != last; ++first) {
+    auto block = std::move(*first);
```
cjh1: Is the `std:move` needed here, wouldn't it use the move semantics anyway?

psavery: Well, the iterator here [returns a reference](https://github.com/OpenChemistry/stempy/blob/master/stempy/reader.h#L98) to the `Block`. I think we could either make a copy of the `Block` or move it. Copying the block is a tad bit more computationally expensive than moving, because copying requires the `shared_ptr` to atomically increment the reference count, while moving doesn't. We've also been [moving it elsewhere](https://github.com/OpenChemistry/stempy/blob/master/stempy/image.cpp#L197).

However, if we pass in a `std::vector<Block>::iterator` instead, we may want to copy, because otherwise we can't use the vector after the function finishes (all of the `Block`s inside will be moved out).

Should I change this one (and the other cases) to copy instead of move?

cjh1: nope, that is good, just wanted to check

PR #27, opened 2019-04-26
Added a benchmark script for electron counting

psavery:
This benchmarks the file I/O, as well as the `calculate_average`
and `electron_count` methods together.

### DIFF
--- a/examples/benchmark_electron_counting.py
+++ b/examples/benchmark_electron_counting.py
@@ -1,0 +1,44 @@
+import click
+from stempy import io, image
+import subprocess
+import sys
+import time
+
+# Beware of disk caching when using this benchmark
+@click.command()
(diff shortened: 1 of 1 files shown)


psavery:
Some quick timings to go with this.

On my laptop, which uses an SSD, has 6 physical cores with hyperthreading (so 12 processes), and has a GeForce MX130 video card, I got these timings for 5 run averages:

```
Serial: 13.95 seconds
OpenMP: 8.23 seconds
CUDA: 13.24 seconds
```

Based on other timings I did, it looks like the I/O time on my laptop for reading the file for `calculateAverage` is 3.27 seconds, and then it is about the same for reading the file for the electron counting. So the total I/O time is around 6.54 seconds. The `calculateAverage` processing time (which is currently always serial) is around 0.29 seconds. The remainder would be electron counting time.

PR #29, opened 2019-04-29
Use a reference when zeroing image numbers

psavery:
Since a reference wasn't being used, a copy was being made,
and then the copy was zeroed. This resulted in the original
image numbers being unchanged, and it resulted in completely
white bright and dark fields.


psavery:
@cjh1 Sorry Chris, I should have tested #25. It unfortunately resulted in all white bright and dark field images. But this fixes the issue.

PR #31, opened 2019-05-03
Add functions to save raw diffraction data to hdf5

psavery:
This does not currently work entirely, because it looks like
the output data has a wrong shape ( 1, 1600, 576, 576 for our
example).


psavery:
@cjh1 Let me know if it looks like I'm on the right track for this. I'm still saving it under the group `electron_events`, even though I guess it's technically not electron events since it is raw data.

cjh1:
@psavery We should probably put it under another group name. @cryos Thoughts?

psavery:
About the shape of the data: I'm actually starting to wonder if that shape is okay. We have one block (which is the reason there is only 1 in the first dimension). There are 1600 scans, and each image is 576 x 576 dimensions.

I can reshape it if we want it in a different shape.

examples/create_hdf5_raw.py
```diff
@@ -0,0 +1,49 @@
...
+    else:
+        sys.exit('Unknown reader version:', reader_version)
+
+    reader = io.reader(files, version=reader_version)
+    blocks = [block for block in reader]
+
+    detector_nx = blocks[0].header.columns
+    detector_ny = blocks[0].header.rows
+
+    raw_data = np.array([block.data for block in blocks], dtype=np.uint32)
```
psavery: Whoops, the datatype here should be uint16 instead (I just noticed that my output data was 2x the size of the input).

psavery:
HDF5 allows storing the dataset in a compressed form, and it's pretty easy to tell h5py how to compress it. I thought we might be interested in compressing the raw data (if it reduces the amount of storage on the server significantly), and I did a few simple tests here with different compression methods and timings.
```
Compression Method        File Size        Compress and Write Time (seconds)
None                      1013M             0.46
gzip (1)                   634M            16.69
gzip (default)             617M            22.62
gzip (9)                   602M            75.51
lzf                        951M             3.58
szip                       614M            16.39
```

Note that gzip has a compression level of 4 by default. I tried the default compression level, a compression level of 1 (the lowest), and a compression level of 9 (the max).

psavery:
I have also been reading more about zipping in HDF5. Apparently, when you zip a dataset, it breaks the dataset up into chunks, and zips each chunk. If you want to read part of the dataset, it only has to decompress the chunks that contain the data you are reading.

If I make the chunk size equal to the diffractogram size, reading individual diffractograms from the dataset is pretty fast, because it only has to decompress one diffractogram. For my computer, reading an individual frame from the compressed dataset only takes ~0.004 seconds. You can see the stats below.
```
Stats for times reading each the 1600 diffractograms one by one:

min was: 0.0031 seconds
max was: 0.0047 seconds
average was: 0.0033 seconds
```

However, it is probably more important to consider the decompression and reading speeds on the computer that will be the server.

cryos:
We chatted a little, and I would keep electron_events for the electron counts as suggested, and use frames at the patch for the raw frames taken without any processing. In the future there might be photon_events. Using events feels a little too generic, and this is a working schema we might change after feedback.

psavery:
Sounds good! I think this is ready for review.

ercius:
A quick comment on compression and chunking in HDF5. Also take note that
the default cache size is only 1MB. This means if a chunk is larger than
the chunk cache then you can really do down reading in some cases. I
suggest using a larger cache size for reading and writing.

psavery:
Thanks for the info, @ercius! I just put in a check here to ensure that, when writing compressed data, the chunk cache size is at least the size of one chunk. 

Although it is an option here to write the raw data in a compressed form, we haven't started working with compressed data yet (we have been sticking to uncompressed). But if we do, we will also need to ensure that the chunk cache size is large enough when reading the file.

For my timings above, it looks like the chunk size was ~0.6 megabytes. So I think the default was okay for them.

PR #35, opened 2019-05-08
Fix compiler warnings

psavery:


psavery:
@cjh1 This is not quite done yet. But the last commit fixes the shadow warnings in constructors.

psavery:
@cjh1 All warnings are gone.

PR #34, opened 2019-05-08
Add support for format 3

cjh1:


python/io.cpp
```diff
@@ -12,8 +12,8 @@ PYBIND11_MODULE(_io, m)
 {
   py::class_<Header>(m, "_header")
     .def_readonly("images_in_block", &Header::imagesInBlock)
-    .def_readonly("rows", &Header::rows)
```
cryos: I don't think it matters lots, but I would assume the default is frame, and not prefix things referring to frames. I would instead prefix things referring to the larger thing, such as stemScanWidth, etc.

cryos:
Looks reasonable to me, I think we can be reasonably relaxed about naming/API with a view to taking a pass when things settle a little more. Thanks for pushing on this Chris.

python/io.cpp
cjh1: I just wanted to make it clear, as it was confusing to me, but we an iterate.

stempy/electron.cpp
```diff
@@ -166,27 +169,51 @@ std::vector<uint32_t> maximalPoints(
...
+  }
+
+  // Raise an exception if we still don't have valid rows and columns
+  if (scanRows <= 0 || scanColumns <= 0) {
+    std::ostringstream msg;
+    msg << "No scan image size provided.";
+    throw std::invalid_argument(msg.str());
+  }
+
+  std::cout << scanRows << std::endl;
```
psavery: Should we eliminate debug messages?

cryos: I don't think it hurts having a few at this stage.

cjh1: I will remove them, I didn't mean to leave them in!

PR #43, opened 2019-05-14
Import pybind11 modules in the new __init__.py

psavery:
Otherwise, `_io` and `_image` will fail to get imported
for local development.


psavery:
We need to check and make sure the docker image still works also.

cjh1:
This doesn't work in the docker image. We will need to rethink this.

Issue #40, opened 2019-05-14
Allow operations such as  create_stem_image to be applied to electron counted data


cjh1:
The approach for now will be to expand the data to the full size and use our current algorithms.

PR #44, opened 2019-05-15
Add ability to calculate stem images from electron counted data

psavery:
This is a work in progress currently.

Fixes: #40

### DIFF
--- a/python/stempy/io/__init__.py
+++ b/python/stempy/io/__init__.py
@@ -37,16 +37,13 @@ class Reader(_reader):
 def reader(path, version=FileVersion.VERSION1):
     return Reader(path, version)
 
-def expand_electron_data(frames, frameRows, frameColumns):
-    res = []
-    for frame in frames:
-        array = np.zeros((frameRows * frameColumns,), dtype=np.uint16)
+def expand_electron_data(frames, numPixels):
+    ret = np.zeros((len(frames) * numPixels,), dtype=np.uint16)
+    for i, frame in enumerate(frames):
         for pos in frame:
-            array[pos] = 1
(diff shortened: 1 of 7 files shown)


psavery:
This might potentially be working. I am able to create some dark and bright field image files from our electron counted data. But they look like static...

dark
bright

python/stempy/io/__init__.py
```diff
@@ -37,6 +37,14 @@ def read(self):
...
     return Reader(path, version)
 
+def expand_electron_data(frames, numPixels):
+    ret = np.zeros((len(frames) * numPixels,), dtype=np.uint16)
+    for i, frame in enumerate(frames):
+        for pos in frame:
+            ret[i * numPixels + pos] = 1
+
+    return ret
+
```
cjh1: Shouldn't we do this expansion in C++, given the looping?

psavery: We can move it to C++, although I don't think this function takes up too much time. ~0.19 seconds in my timing (the STEM calculation is taking around 50 seconds).

cjh1: ok, that is fine, leave for the moment

cjh1: 50 seconds on the electron counted data?

psavery:
The latest commit results in a very dramatic speed increase, from
~50 - 60 seconds on my computer to 0.5 seconds.

Timings seem to indicate that the vast majority of time
was spent in between the pybind11 function call and the start
of the C++ function. Whatever was happening in between (perhaps
a data copy or conversion) was taking a long time.

It is now much faster.

Issue #39, opened 2019-05-14
Allow the center of annular mask to be specified


cjh1:
This would be optional, default to the center. The idea is that is would be able to be provided on [this call](https://github.com/OpenChemistry/stempy/blob/master/python/stempy/image/__init__.py#L5).

PR #45, opened 2019-05-15
Allow center of annular masks to be specified

psavery:
If the values for the centers are less than 0, they
will be automatically calculated.

Fixes: #39

For our stem image sample, the default, 288, is this:

Bright
=====
bright_original

Dark
====
dark_original

If we change it to 250, for example:

Bright
=====
bright

Dark
====
dark

### DIFF
--- a/stempy/mask.cpp
+++ b/stempy/mask.cpp
@@ -7,11 +7,18 @@ using namespace std;
 
 namespace stempy {
 
-uint16_t* createAnnularMask(int rows, int columns, int innerRadius, int outerRadius) {
+uint16_t* createAnnularMask(int rows, int columns, int innerRadius,
+                            int outerRadius, int centerX, int centerY)
+{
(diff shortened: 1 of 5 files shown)


stempy/mask.cpp
```diff
@@ -7,19 +7,26 @@ using namespace std;
...
   auto numberOfElements = rows*columns;
   auto mask = new uint16_t[numberOfElements]();
-  auto xCenter = round(rows/2.0);
-  auto yCenter = round(rows/2.0);
+
+  if (centerX < 0)
+    centerX = round(rows / 2.0);
+
+  if (centerY < 0)
+    centerY = round(rows / 2.0);
```
psavery: Should one of these be calculated via the columns instead of the rows?

cjh1: yes

PR #50, opened 2019-05-23
Added ability to create many stem images in one go

psavery:
Multiple stem images can now be created in one go, which can
save a lot of time compared to doing them all individually.

A couple of python examples were modified to illustrate the new API.
But the old python API still works for backward compatibility.

### DIFF
--- a/examples/electron_stem_image.py
+++ b/examples/electron_stem_image.py
@@ -33,4 +33,4 @@ img = image.create_stem_image_sparse(frames, inner_radius, outer_radius,
                                      scan_width, scan_height, frame_width,
                                      frame_height)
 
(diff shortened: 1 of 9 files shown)


psavery:
A major part of this PR as well is making it so that we do not, by default, create bright and dark field images. We just create whatever images the user asks for, and the user does so by specifying the inner_radii and outer_radii.

For our stem image examples, the way to obtain the equivalent bright and dark field images is by using `inner_radii = [0, 40]` and `outer_radii = [288, 288]`. The first of the resulting images will be the equivalent of our sample bright field image, and the second will be the equivalent of our sample dark field image.

PR #51, opened 2019-05-23
Modify save_stem_images() to allow compatibility

psavery:
Modify save_stem_images() so that the images may be saved
as `bright` and `dark` just as they were before. Now, a
list of images is passed along with a list of names.


psavery:
@cjh1 Can we do this so that we do not need to modify the stem_server API?

cjh1:
@psavery I Like the idea of being able to the 'name' the images, but I think we want to be able to access arbitrary images from the file, not just "bright' and "dark". May be we could preserve the ability to have names by having another dataset to store the names in making indexes?

PR #55, opened 2019-05-28
Ensure thread safety

cjh1:


python/image.cpp
```diff
@@ -41,7 +41,7 @@ PYBIND11_MODULE(_image, m)
     .def_buffer([](RadialSum<uint64_t>& r) {
        return py::buffer_info(
           r.data.get(),                                                 /* Pointer to buffer */
-          sizeof(double),                                               /* Size of one scalar */
+          sizeof(uint64_t),                                               /* Size of one scalar */
           py::format_descriptor<double>::format(),                    /* Python struct-style format descriptor */
```
psavery: This line uses the `double` also.

cjh1: fixed

PR #56, opened 2019-05-29
Add VTK-m implementation of radial sum

cjh1:


stempy/image.cpp
```diff
@@ -386,11 +395,76 @@ void radialSumFrame(int centerX, int centerY, const uint16_t data[],
...
+                   radialSum);
+  }
+}
+
+#else
+void radialSumFrames(int centerX, int centerY, const uint16_t data[],
+                     int frameWidth, int frameHeight,
+                     std::vector<uint32_t>& imageNumbers,
+                     uint32_t numberOfPixels,
+                     uint32_t numberOfScanPositions RadialSum<uint64_t>& radialSum)
```
psavery: Haven't tried compiling these yet, but it looks like a comma is missing here.

stempy/image.cpp
```diff
@@ -386,11 +395,76 @@ void radialSumFrame(int centerX, int centerY, const uint16_t data[],
...
+  using ControlSignature = void(FieldIn, AtomicArrayInOut);
+
+  using ExecutionSignature = void(_1, _2, WorkIndex);
+
+  template <typename AtomicInterface>
+  VTKM_EXEC void operator()(const vtkm::UInt16& value,
+                            AtomicInterface& radialSum, vtkm::Id i) const
+  {
+    auto x = i % m_shape[0];
+    auto y = i / m_shape[0];
```
psavery: Is `m_shape[1]` supposed to be used anywhere? I haven't seen it being used.

stempy/image.cpp
```diff
@@ -457,6 +533,13 @@ RadialSum<uint64_t> radialSum(InputIt first, InputIt last, int scanWidth, int sc
   ThreadPool pool(numThreads);
   RadialSum<uint64_t> radialSum(scanWidth, scanHeight, maxRadius+1);
 
+#ifdef VTKm
+  // We need the reinterpret_cast as vtkm currently doesn't support atomic
+  // access for uint64.
+  auto radialSumHandle = vtkm::cont::make_ArrayHandle(
+    reinterpret_cast<vtkm::Int64*>(radialSum.data.get()),
```
psavery: Should this be `vtkm::UInt64*` instead?

psavery: Oh okay, I see your comment above now.

stempy/image.cpp
psavery: Indeed, I tried compiling this with VTKm turned off. I get a compilation error here, and I think one or two others. Might be good to check them out.

stempy/image.cpp
cjh1: Nope, I guess I can just pass in the width

stempy/image.cpp
cjh1: Will do

cjh1: Done

stempy/image.cpp
cjh1: Done

PR #57, opened 2019-06-04
Update the Stempy H5 Format doc

psavery:
This reflects recent changes to the format.

### DIFF
--- a/docs/StempyH5Format.md
+++ b/docs/StempyH5Format.md
@@ -10,24 +10,27 @@ electron_events
 ├── frames
 └── scan_positions
 stem
-├── bright
-└── dark
+└── images
+frames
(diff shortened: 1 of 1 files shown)


docs/StempyH5Format.md
```diff
@@ -10,24 +10,27 @@ electron_events
 ├── frames
 └── scan_positions
 stem
-├── bright
-└── dark
+└── images
+frames
```
cjh1: Is this `frames` left over?

psavery: That's the root `/frames`, which is the raw diffractogram data.

cjh1: ok,make sense

PR #59, opened 2019-06-20
Pull request from Zhuokai

zhuokaizhao:
In general, I added Boost in CMakeList.txt for file system support, and histogram stuff in equalizer.h and equalizer.cpp. 
Minor changes were made in image.cpp.
main.cpp now runs the histogram.
Please feel free to let me know if there were bugs, thank you!

Zhuokai


CMakeLists.txt
```diff
@@ -32,6 +32,10 @@ include_directories(
 find_package(Eigen3 REQUIRED)
 include_directories(${EIGEN3_INCLUDE_DIR})
 
+# add boost support for file system
+find_package(Boost REQUIRED COMPONENTS filesystem system)
```
cryos: I would like to avoid any dependency on Boost

zhuokaizhao: Sure, I will make the change to use std::filesystem instead. Can I change "set(CMAKE_CXX_STANDARD 14)" to 17?

cryos: We will have to take a look at whether that is realistic for where we want to deploy it. We can talk more about what you need, and as this only needs to work on Linux we may just do something less portable in the short term.

stempy/electron.cpp
```diff
@@ -114,51 +114,61 @@ std::vector<uint32_t> maximalPointsParallel(
...
-    event = event &&
-            pixelValue > frame[bottomNeighbourRowIndex + leftNeighbourColumn];
+    // if larger than bottom left neighbor pixel value
+    event = event && pixelValue > frame[bottomNeighbourRowIndex + leftNeighbourColumn];
     // left
     event = event && pixelValue > frame[rowIndex + leftNeighbourColumn];
     // top left
-    event =
-      event && pixelValue > frame[topNeighbourRowIndex + leftNeighbourColumn];
+    event = event && pixelValue > frame[topNeighbourRowIndex + leftNeighbourColumn];
```
psavery: The reason these are split over two lines is because we impose an 80 character limit per line. These are probably going to need to be changed back to how they were...

You can see our style rules in the [.clang-format file](https://github.com/OpenChemistry/stempy/blob/master/.clang-format). And you can use the `clang-format` program to automatically format the files for the style rules.

stempy/electron.cpp
```diff
@@ -114,51 +114,61 @@ std::vector<uint32_t> maximalPointsParallel(
...
     auto bottomNeighbourRow = mod((row + 1), height);
+
+    // pixel value at the current pixel
     auto pixelValue = frame[i];
+
+    // start index of the first element in a row
     auto bottomNeighbourRowIndex = bottomNeighbourRow * width;
     auto topNeighbourRowIndex = topNeighbourRow * width;
+
+    // is rowIndex the same as i?
```
psavery: I'm not sure if we should be committing lots of extra comments to the code. These look like they might better belong in a personal branch rather than in the master branch that everyone uses.

python/equalizer.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include <vector>
+
+#include <stempy/image.h>
+#include <stempy/equalizer.h>
+
+namespace py = pybind11;
+
+using namespace stempy;
+
+PYBIND11_MODULE(_image, m)
```
psavery: This pybind11 module probably needs to be named something other than `_image`, as that name is already taken. It also looks like there might be several header files included that are not needed.

CMakeLists.txt
psavery: Looks like he is mostly using it in main.cpp so that you can run the histogram stuff from the C++ executable.

@zhuokaizhao, we have mostly been using python and pybind11 to run the underlying C++ functions rather than running them in the `main()` C++ function. Most of our examples are either [ipython-notebooks](https://github.com/OpenChemistry/stempy/blob/master/examples/stem_image.ipynb) or [python scripts](https://github.com/OpenChemistry/stempy/blob/master/examples/electron_stem_image.py). It might be good for you to write a histogram example in python rather than in C++, so it is easier to deal with filesystems.

stempy/main.cpp
```diff
@@ -12,24 +64,98 @@ int main (int argc, char *argv[])
...
+
+      // generate current partial STEM image
+      cout << "Generating partial STEM image" << endl;
+      
+      // information about how to process
+      vector<int> innerRadii(stream.header.imagesInBlock, 40);
+      vector<int> outerRadii(stream.header.imagesInBlock, 288);
+      
+      // create STEM 32 partial STEM images
+      vector<STEMImage> imagesOfBlock = createSTEMImages(reader.begin(), reader.end(), innerRadii, outerRadii, width, height, centerX, centerY);
```
psavery: Correct me if I'm wrong, but it looks like this is producing several identical stem images? It might be faster to create one, and then copy it to get the rest of them, although you would have to make sure the `shared_ptr`'s inside get deep copied.

CMakeLists.txt
cjh1: I think @zhuokaizhao was using main.cpp for testing the algorithm, now that we have the python binds in place, we should just add a notebook or script, we can do any filesystem things there.

python/equalizer.cpp
cjh1: We can probably just add this to the _image module, rather than having a separate module.

stempy/electron.cpp
cjh1: This is a [nice script](https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/git-clang-format) that will integrate with git, so you can run `git clang-format`

stempy/equalizer.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <iostream>
+
+
+namespace stempy 
+{
+    // hide helper functions
+    namespace 
+    {
+        // helper function that returns the binding result
+        std::vector<float> getcontainer(const STEMImage inImage, const int numBins)
```
cjh1: We like to use camelCase for method/function names.

stempy/equalizer.cpp
```diff
@@ -0,0 +1,104 @@
...
+            // }
+            // std::cout << std::endl;
+
+            return container;
+        }
+    }
+    
+    // function that computes histogram for all the STEM images
+    // each histogram is a vector<int>
+    std::vector<int> createSTEMHistograms (STEMImage inImage, const int numBins)
```
cjh1: Isn't this just create a histogram for a singel STEM image, in which case it should be `createSTEMHistogram`?

stempy/image.cpp
```diff
@@ -28,13 +28,24 @@ using namespace std;
...
 }
 
+// copy constructpr
+template <typename T>
+Image<T>::Image(const Image& image) 
+{
+  width = image.width; 
+  height = image.height;
+  data = image.data; 
+}
```
cjh1: Why do we need a copy constructor? We were trying to rely on using move semantics.

stempy/image.h
```diff
@@ -14,10 +14,14 @@ namespace stempy {
     uint32_t height = 0;
     std::shared_ptr<T[]> data;
 
+    // default constructor
     Image() = default;
+    // constructor
     Image(uint32_t width, uint32_t height);
-    Image(Image&& i) noexcept = default;
+    // Image(Image&& i) noexcept = default;
```
cjh1: We what this move constructor.

stempy/main.cpp
```diff
@@ -12,24 +64,98 @@ int main (int argc, char *argv[])
...
+    int centerX = -1;
+    int centerY = -1;
+
+    for (int i = 0; i < numFiles; i++)
+    {
+      cout << "Processing " << i+1 << "/" << numFiles << ": " << files[i] << endl;
+      string curFile = files[i];
+      // read the data
+      stempy::StreamReader reader(inputPath+curFile);
+      auto stream = reader.read();
```
psavery: Calling `read()` on the `StreamReader` reads one block and increments the underlying `ifstream`. You might need to call `reader.reset()` to reset it to the beginning if you are going to use it again later.

stempy/equalizer.cpp
psavery: We are normally using `camelCase` for stempy, so `getContainer` would be a better name, I think.

stempy/equalizer.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include <vector>
+
+namespace stempy {
+
+    // function that computes histogram for all the STEM images
+    std::vector<int> createSTEMHistograms (STEMImage inImage, const int numBins);
+}
+
+#endif
```
cjh1: Think we can just add the `createSTEMHIstogram` to `image.h` and `image.cpp`, rather than having `equalizer.h` and `equalizer.cpp`

stempy/image.cpp
psavery: This looks like it does the same thing as the default copy constructor anyways. We can probably leave it out.

stempy/electron.cpp
psavery: Yeah! `git-clang-format` is great.

stempy/electron.cpp
zhuokaizhao: I am so sorry that it was just my personal habit of writing random useless comments while reading other people's code. You are right that this indeed should belong to a personal branch. I apologize again for forgetting ignoring these while making the pull request.

python/equalizer.cpp
zhuokaizhao: @psavery I actually haven't really got into the pybind11 step yet. It was just there but it is not finished yet. 
@cjh1 Sure, I will look into a way to add it to the _image module, thank you for your suggestion.

stempy/equalizer.cpp
zhuokaizhao: Sure, I will change it to getContainer, thank you.

stempy/electron.cpp
zhuokaizhao: Thanks! I've installed clang-format, I am sorry that I did not know this.

stempy/equalizer.cpp
zhuokaizhao: Oh yeah definitely, I forgot to change the name after our yesterday's discussion.

stempy/main.cpp
zhuokaizhao: Yeah you are correct, sorry I misunderstood what createSTEMImages() returned. Thank you very much!

stempy/image.cpp
zhuokaizhao: Thanks, initially I was trying to do something like "for (singleSTEMImage : allSTEMImages)" and it gave me error that I thought was because of no copy constructor. But later I realized that I did not even need to have the for loop. But I forgot to delete this copy constructor.

stempy/equalizer.h
zhuokaizhao: I was implementing it in the new files because I was thinking later it might also contain equalizer stuff (thus the naming). But we can definitely put all those stuff within image.h and image.cpp if that makes more sense to you.

stempy/main.cpp
zhuokaizhao: I might be wrong but I think that for each iteration, reader is newly created and gets a new file, instead of getting a single block. In this case, do we still need to reset it?

stempy/equalizer.h
cjh1: My thought is that the histogram creation functionality will be useful not just for the contrast equalization, so probably belongs with the other image processing functions.

zhuokaizhao: Sure, I will then integrate equalizer.h/cpp into image.h/cpp.

stempy/main.cpp
psavery: Yes, that is true. But I was referring more to [this line](https://github.com/OpenChemistry/stempy/pull/59/files#diff-f87a2c458e9afaf9bbf1d223829b9e4dR130) below where the reader gets re-used again. You might need to call `reader.reset()` before that line, because the `StreamReader::begin()` function doesn't actually reset the `ifstream` position, but starts reading from wherever it currently is.

It might be more intuitive for us to get `StreamReader::begin()` to start from the beginning, though. What do you think, @cjh1?

psavery: (Note that `StreamReader::begin()` is [here](https://github.com/OpenChemistry/stempy/blob/master/stempy/reader.h#L55))

stempy/electron.cpp
psavery: No problem. `git-clang-format` is particularly good because it will only run clang-format on your changes (rather than running clang-format on the whole files).

stempy/image.cpp
```diff
@@ -578,6 +592,81 @@ RadialSum<uint64_t> radialSum(InputIt first, InputIt last, int scanWidth, int sc
...
+            << ", height = " << height << std::endl;
+
+  // curImage.data is shared_ptr<uint64_t []>
+  for (int i = 0; i < width * height; i++) {
+    // std::cout << "curData[" << i << "] = " << curData[i] << std::endl;
+    if (curData[i] < min)
+      min = curData[i];
+    if (curData[i] > max)
+      max = curData[i];
+  }
```
psavery: By the way, I don't think this needs to be changed for this commit, but for future reference, it's good to use STL algorithms where possible, such as `std::minmax_element`. I believe that could be used here like so:
```
auto result = std::minmax_element(curData.get(), curData.get() + width * height);
uint64_t min = *result.first;
uint64_t min = *result.second;
```

stempy/image.cpp
```diff
@@ -578,6 +592,81 @@ RadialSum<uint64_t> radialSum(InputIt first, InputIt last, int scanWidth, int sc
...
+
+  // the "length" of each slot of the container
+  float length = (float)((max - min) / numBins);
```
psavery: Also, we usually try to use C++-style casts (`static_cast<float>()`) rather than C-style casts.

zhuokaizhao: Thank you for pointing it out, I have fixed it.

stempy/image.cpp
zhuokaizhao: Thank you very much for including the actual code, I've changed it to STL algorithm.

PR #61, opened 2019-06-25
Added feature that creates histogram of STEM images

zhuokaizhao:
Example program is stem_histogram.py. It can be run by 
`python3 examples/stem_histogram.py <data_dir> <output_dir>`

Sample results look like
histogram_0_288
histogram_40_288

Please ignore the changes made in main.cpp. It was only used for early-stage testing. Thank you for reviewing.


cryos:
Before this is ready to be reviewed you should remove any changes you don't want to merge, such as those in main.cpp. We also want to get rid of all of the old commits, and create a new PR with one or two commits that describe/show the feature you are adding only. You can use [interactive rebasing](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) to squash them together, or transplant changes that make sense to keep.

PR #62, opened 2019-06-25
Added a feature that computes histograms of STEM images

zhuokaizhao:
Example program is stem_histogram.py. It can be run by
`python3 examples/stem_histogram.py <data_dir> <output_dir>`

Sample results look like
histogram_0_288
histogram_40_288

Thank you for reviewing.

### DIFF
--- a/examples/benchmark_stem_image.py
+++ b/examples/benchmark_stem_image.py
@@ -2,6 +2,7 @@ import click
 from stempy import io, image
 import sys
 import time
+import numpy as np
 
 # Beware of disk caching when using this benchmark
 @click.command()
@@ -23,6 +24,7 @@ def run_benchmarks(files, num_runs):
 
         reader = io.reader(files)
         img = image.create_stem_image(reader, 40, 288, width=160, height=160)
+        img = np.array(img, copy=False)
 
         end = time.time()
         times.append(end - start)
--- a/examples/benchmark_stem_image.py
+++ b/examples/benchmark_stem_image.py
@@ -2,7 +2,6 @@ import click
 from stempy import io, image
 import sys
 import time
-import numpy as np
 
 # Beware of disk caching when using this benchmark
 @click.command()
--- a/examples/benchmark_stem_image.py
+++ b/examples/benchmark_stem_image.py
@@ -24,7 +24,6 @@ def run_benchmarks(files, num_runs):
 
         reader = io.reader(files)
         img = image.create_stem_image(reader, 40, 288, width=160, height=160)
-        img = np.array(img, copy=False)
 
         end = time.time()
         times.append(end - start)
--- a/examples/stem_histogram.py
+++ b/examples/stem_histogram.py
@@ -1,0 +1,91 @@
+import glob
+import sys
+import os
+from stempy import io, image
+import numpy as np
+from PIL import Image
+import matplotlib.pyplot as plt
+import matplotlib.ticker as ticker
+
+def print_help():
+    print(
+    '''
+    Usage:
+    python3 stem_histogram.py [dataDir -- path to the data directory]
+                              [outDir -- path to the output image directory]
+    ''')
+
+def save_img(stem_image_data, name):
+    min = np.min(stem_image_data)
+    max = np.max(stem_image_data)
+
+    stem_image_data = stem_image_data.reshape((160, 160))
+    stem_image_data = np.interp(stem_image_data, [min, max], [0, 256])
+    stem_image_data = stem_image_data.astype(np.uint8)
+    img = Image.fromarray(stem_image_data)
+    img.save(name)
+
+
+def main(argv):
+    print_help()
+    dataDir = argv[0] if len(argv) > 0 else '/home/zhuokai/Desktop/data/data*.dat'
+    outDir = argv[1] if len(argv) > 1 else '/home/zhuokai/Desktop/stempy/examples/output/'
+
+    # create output directory if it does not exit
+    if not os.path.exists(outDir):
+        os.mkdir(outDir)
+        print('Output directory', outDir , 'created')
+    else:    
+        print('Outout directory', outDir , 'already exists')
+
+    # get the all the data files
+    files = []
+    for f in glob.glob(dataDir):
+        files.append(f)
+
+    # inner and outer radius of mask
+    inner_radii = [0, 40]
+    outer_radii = [288, 288]
+
+    reader = io.reader(files)
+    imgs = image.create_stem_images(reader, inner_radii, outer_radii, width=160,
+                                    height=160)
+
+    # save the STEM images
+    for i, img in enumerate(imgs):
+        # convert the type to numpy array
+        img_np = np.array(img, copy=False)
+        suffix = str(inner_radii[i]) + '_' + str(outer_radii[i]) + '.png'
+        save_img(img_np, outDir + '/img_' + suffix)
+        print('STEM image with inner radius = ' + str(inner_radii[i]) 
+              + ', outer radius = ' + str(outer_radii[i]) + ' has been saved')
+
+    # generate histograms
+    numBins = 100
+    for i, img in enumerate(imgs):
+        print('Generating histogram for STEM image ' + str(i))
+        bins, freq = image.create_stem_histogram(img, numBins)
+        bins = [str(element) for element in bins]
+
+        # plot histogram
+        fig = plt.figure(1, figsize=(16, 8))
+        myHist = fig.add_subplot(111)
+        # plt.bar considers the left boundary
+        x = np.arange(numBins+1)
+        myHist.bar(x[:-1], freq, align='edge')
+        plt.xticks(x[::20], bins[::20])
+        plt.title('Histogram of STEM image with inner radius = ' 
+                    + str(inner_radii[i]) + ', outer radius = ' + str(outer_radii[i]))
+        plt.xlabel('Value')
+        plt.ylabel('Frequency')
+
+        # save to local
+        suffix = str(inner_radii[i]) + '_' + str(outer_radii[i]) + '.png'
+        plt.savefig(outDir + '/histogram_' + suffix)
+        
+        plt.show()
+
+
+if __name__ == "__main__":
+    main(sys.argv[1:])
+
--- a/examples/stem_histogram.py
+++ b/examples/stem_histogram.py
@@ -1,3 +1,4 @@
+import click
 import glob
 import sys
 import os
@@ -25,11 +26,10 @@ def save_img(stem_image_data, name):
     img = Image.fromarray(stem_image_data)
     img.save(name)
 
-
 def main(argv):
     print_help()
-    dataDir = argv[0] if len(argv) > 0 else '/home/zhuokai/Desktop/data/data*.dat'
-    outDir = argv[1] if len(argv) > 1 else '/home/zhuokai/Desktop/stempy/examples/output/'
+    dataDir = argv[0] if len(argv) > 0
+    outDir = argv[1] if len(argv) > 1
 
     # create output directory if it does not exit
     if not os.path.exists(outDir):
@@ -44,30 +44,24 @@ def main(argv):
         files.append(f)
 
     # inner and outer radius of mask
-    inner_radii = [40]
-    outer_radii = [288]
+    inner_radii = [0, 40]
+    outer_radii = [288, 288]
 
+    # file reader
     reader = io.reader(files)
-    imgs = image.create_stem_images(reader, inner_radii, outer_radii, width=160,
-                                    height=160)
-
-    # save the STEM images
-    for i, img in enumerate(imgs):
-        # convert the type to numpy array
-        img_np = np.array(img, copy=False)
-        suffix = str(inner_radii[i]) + '_' + str(outer_radii[i]) + '.png'
-        save_img(img_np, outDir + '/img_' + suffix)
-        print('STEM image with inner radius = ' + str(inner_radii[i]) 
-              + ', outer radius = ' + str(outer_radii[i]) + ' has been saved')
 
     # generate histograms
     numBins = 100
-    for i, img in enumerate(imgs):
-        print('Generating histogram for STEM image ' + str(i))
-        bins, freq = image.create_stem_histogram(img, numBins)
-        bins = [str(element) for element in bins]
-
-        # plot histogram
+    print('Generating histograms for input data')
+    allBins, allFreqs = image.create_stem_histogram(numBins, reader, inner_radii, outer_radii,
+                                                    width=160, height=160)
+            
+    # plot histogram
+    for i in range(len(allBins)):
+        # obtain current bins and freq
+        bins = [str(element) for element in allBins[i]]
+        freq = allFreqs[i]
+        # init figure
         fig = plt.figure(1, figsize=(16, 8))
         myHist = fig.add_subplot(111)
         # plt.bar considers the left boundary
--- a/examples/stem_histogram.py
+++ b/examples/stem_histogram.py
@@ -1,4 +1,3 @@
-import glob
 import sys
 import os
 from stempy import io, image
@@ -27,20 +26,21 @@ def save_img(stem_image_data, name):
 
 def main(argv):
     print_help()
-    dataDir = argv[0] if len(argv) > 0
-    outDir = argv[1] if len(argv) > 1
+    dataDir = argv[0]
+    outDir = argv[1]
 
     # create output directory if it does not exit
     if not os.path.exists(outDir):
         os.mkdir(outDir)
         print('Output directory', outDir , 'created')
-    else:    
-        print('Outout directory', outDir , 'already exists')
+    else:
+        print('Output directory', outDir , 'already exists')
 
     # get the all the data files
     files = []
-    for f in glob.glob(dataDir):
-        files.append(f)
+    for root, dirs, fs in os.walk(dataDir):
+        for f in fs:
+            files.append(os.path.join(root, f))
 
     # inner and outer radius of mask
     inner_radii = [0, 40]
@@ -54,7 +54,7 @@ def main(argv):
     print('Generating histograms for input data')
     allBins, allFreqs = image.create_stem_histogram(numBins, reader, inner_radii, outer_radii,
                                                     width=160, height=160)
-            
+
     # plot histogram
     for i in range(len(allBins)):
         # obtain current bins and freq
@@ -67,7 +67,7 @@ def main(argv):
         x = np.arange(numBins+1)
         myHist.bar(x[:-1], freq, align='edge')
         plt.xticks(x[::20], bins[::20])
-        plt.title('Histogram of STEM image with inner radius = ' 
+        plt.title('Histogram of STEM image with inner radius = '
                     + str(inner_radii[i]) + ', outer radius = ' + str(outer_radii[i]))
         plt.xlabel('Value')
         plt.ylabel('Frequency')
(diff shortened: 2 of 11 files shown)


stempy/image.cpp
```diff
@@ -269,6 +270,62 @@ vector<STEMImage> createSTEMImages(InputIt first, InputIt last,
...
 }
 
+// returns the binding result
+std::vector<double> getContainer(const STEMImage& inImage, const int numBins)
+{
+  // information about input STEMImage
+  int width = inImage.width;
+  int height = inImage.height;
+  auto curData = inImage.data;
+  std::cout << "Input STEM Image has width = " << width
```
cryos: Might be useful while developing, but this should be removed before we merge.

stempy/image.cpp
```diff
@@ -269,6 +270,62 @@ vector<STEMImage> createSTEMImages(InputIt first, InputIt last,
...
+            << ", height = " << height << std::endl;
+
+  auto result = std::minmax_element(curData.get(), curData.get() + width * height);
+  double min = *result.first;
+  double max = *result.second;
+
+  // the "length" of each slot of the container
+  double length = static_cast<double>((max - min) / numBins);
```
cryos: The way that you have implemented this, you are using the bins as a lower and upper bound on the value to see if it should be added to the bin. In that case in order to properly assign upper and lower bounds you should use a length of `(max - min) / (numBins - 1)`. That would avoid the necessity in the following function of adding max to the last slot - they should all be in range.

Example, I want four bins between 0 and 1, that would result in 0.0, 0.25, 0.50 and 0.75 using the logic above. If I used `numBins - 1` it would yield 0.0, 0.33, 0.66 and 1.0 fully covering the intended range.

stempy/image.cpp
```diff
@@ -269,6 +270,62 @@ vector<STEMImage> createSTEMImages(InputIt first, InputIt last,
...
+  // get a histrogram
+  for (int i = 0; i < width * height; i++) {
+    auto value = curData[i];
+    // check which bin it belongs to
+    for (int i = 0; i < numBins; i++) {
+      if (value >= bins[i] && value < bins[i + 1]) {
+        frequencies[i] += 1;
+      }
+      // the max value is put in the last slot
+      else if (value == bins[numBins]) {
```
cryos: This shouldn't be necessary with the above suggested change to the binning.

stempy/image.cpp
```diff
@@ -269,6 +270,62 @@ vector<STEMImage> createSTEMImages(InputIt first, InputIt last,
...
+  int height = inImage.height;
+  auto curData = inImage.data;
+  
+  // get a histrogram
+  for (int i = 0; i < width * height; i++) {
+    auto value = curData[i];
+    // check which bin it belongs to
+    for (int i = 0; i < numBins; i++) {
+      if (value >= bins[i] && value < bins[i + 1]) {
+        frequencies[i] += 1;
```
cryos: It is potentially a little faster, and just as clear to pre-increment, i.e. `++frequencies[i];`

stempy/image.cpp
cryos: As max and min are already double I don't think the cast is needed either.

zhuokaizhao: Continue with your example, after adding max to my current method, returned `container` would include 4 intervals/bins; [0.0, 0.25), [0.25, 0.5), [0.5, 0.75) and [0.75, 1.0], which matches with user input `numBins=4`. 
The method you suggested does cover the intended range, but only generate 3 bins, which are [0.0, 0.33), [0.33, 0.66) and [0.66, 1.0). Did I miss anything here?

zhuokaizhao: I was unsure if `numBins` being an `int` would cause any rounding errors, but if are certain about this, I can definitely remove the extra cast. Thanks!

stempy/image.cpp
zhuokaizhao: I have removed the extra printouts, thank you!

stempy/image.cpp
zhuokaizhao: Thank you, I have made the change.

stempy/image.cpp
cryos: Your code also looks like it needs `bins + 1`, how is it not walking off the end of the array? See the conditional below, using `value >= bins[i] && value < bins[i + 1]` where `i < numBins`. No rounding error would occur, if you want to force double precision (not needed here) you would need to cast `numBins`.

stempy/image.cpp
```diff
@@ -269,6 +270,62 @@ vector<STEMImage> createSTEMImages(InputIt first, InputIt last,
...
+  double min = *result.first;
+  double max = *result.second;
+
+  // the "length" of each slot of the container
+  double length = static_cast<double>((max - min) / numBins);
+
+  std::vector<double> container;
+  // push all the intermediate values
+  for (int i = 0; i <= numBins; i++) {
+    container.push_back(min + i * length);
```
cryos: I see here, I missed this where you use `<=` numBins, so you should cover the complete range as you have the extra bound at the top.

stempy/image.cpp
cryos: My question becomes - are you sure this is necessary? I guess you are trying to cover the corner case as it is `<` rather than `<=`. It is a little odd to see the spike at the end of the histograms you showed too, is that real, I wouldn't expect it to be.

stempy/image.cpp
zhuokaizhao: When generating the container, my for loop is `for (int i = 0; i **<=** numBins; i++)`, so continue with your example, it would be 5 numbers; 0, 0.25, 0.5, 0.75 and 1 after the loop. 
However, when generating frequencies, the for loop is `for (int i = 0; i **<** numBins; i++)`, which is why it is not walking off the end of the array.

stempy/image.cpp
zhuokaizhao: I think it is necessary. Although it would not give you bugs if I just change the interval to be both-side included, I feel like it produces uncertainty to values happen to be equal to the boundary value. Because technically it can belong to either interval. A clear classification makes me feel a bit more comfortable. But I agree that it is a little odd in the end to separately include the max-corner-case.

python/stempy/image/__init__.py
```diff
@@ -8,8 +8,15 @@ def create_stem_images(reader, inner_radii,
                                      inner_radii, outer_radii, width, height,
                                      center_x, center_y)
 
-    images = [np.array(img, copy=False) for img in imgs]
-    return np.array(images, copy=False)
+    # images = [np.array(img, copy=False) for img in imgs]
+    # return np.array(images, copy=False)
+    return imgs
```
psavery: This change breaks the `stem_image.py` example, and possibly some other scripts as well. It'd probably be preferable to not change this function. Or, if it is needed, modify all the other scripts so that they work with the changes.

psavery: Looks like it breaks the `create_hdf5.py` example also.

stempy/image.cpp
zhuokaizhao: I have fixed the bug, the max case should be outside the for loop which determines which bin the current value it belongs to. The weird artifact was due to the max being added 100 times.

zhuokaizhao: Thank you very much for catching the bug! @cryos

python/stempy/image/__init__.py
zhuokaizhao: I am sorry that I forgot to modify other example scripts. 

I think the change is beneficial - not only to my function which takes STEMImage as input type, but also would make future post-processings easier. I personally think that, since everything is computed through type `STEMImage` in C++, Numpy array is only useful when saving or displaying result in Python, and should be an "extra" step in Python rather than wrapping inside a function.

zhuokaizhao:
Thanks to @cryos who pointed out the weird artifacts in the initial histograms, bug that added max-corner case multiple times has been fixed. The updated histograms look like
histogram_0_288
histogram_40_288

python/stempy/image/__init__.py
zhuokaizhao: I have made the corresponding changes in example scripts that include function `create_stem_images()`. 
Additional lines of either (depends on if single or multiple STEM images have been returned)
`img=np.array(img, copy=False)` 
or 
`imgs = [np.array(img, copy=False) for img in imgs]`
`imgs = np.array(imgs, copy=False)`
has been added.

psavery: Okay, sounds good. Thank you!

examples/stem_histogram.py
```diff
@@ -0,0 +1,91 @@
...
+    stem_image_data = np.interp(stem_image_data, [min, max], [0, 256])
+    stem_image_data = stem_image_data.astype(np.uint8)
+    img = Image.fromarray(stem_image_data)
+    img.save(name)
+
+
+def main(argv):
+    print_help()
+    dataDir = argv[0] if len(argv) > 0 else '/home/zhuokai/Desktop/data/data*.dat'
+    outDir = argv[1] if len(argv) > 1 else '/home/zhuokai/Desktop/stempy/examples/output/'
```
psavery: It's probably okay to leave the argument parsing here as-is for now. But if it gets any more complicated in the future, I would recommend using [click](https://click.palletsprojects.com/en/7.x/) as we have done in some of the other [sample files](https://github.com/OpenChemistry/stempy/blob/1f4eb12cdbf8a845bd9e849e6a6a72d5ad1b09df/examples/create_hdf5.py#L7). It's a nice argument parsing library, and it's already a dependency of stempy.

cryos: I think it would be best to remove hard-coded paths from our code, keep it locally if you like, but please remove it.

python/stempy/image/__init__.py
cryos: Why not just create another function? You are pushing more boiler plate code to the outside, when you could create a new wrapper with the old name that simply does this conversion to a NumPy array. Making an additional function that returns the imgs you want, and then having a simple forwarding function that returns a NumPy array would give us the best of both worlds. Users from the Python side will almost always want the NumPy array, and should get it from our API.

cryos:
I am glad you were able to track down the bug in the histogram maximum value. I think you should revert the changes to the scripts, and have the default public function return a numpy array to avoid pushing all that boilerplate into scripts/notebooks, and add an additional intermediate function if you would like the C++ object for further processing. I want to maintain a focus on an intuitive Python interface even if it adds a little complexity.

stempy/image.cpp
psavery: As a side note on this: notice as well that a direct comparison is being made between a `uint64_t` (`value`) and a `double` (`bins[numBins]` - whose value is a result of arithmetic). I am not sure if this would be an issue or not, but thought I would point it out.

python/stempy/image/__init__.py
zhuokaizhao: @cryos That indeed sounds like a better idea! I'm working on it right now, thanks.

examples/stem_histogram.py
zhuokaizhao: @psavery Thanks, I've changed the current code to use click. @cryos Sure, I have removed the paths.

zhuokaizhao:
I have reverted the previous changes that I made to other example scripts, and have removed the hard-coded paths. I did not change to click in the end since there are just two arguments. Please feel free to let me know if there are other changes needed. Thanks everyone.

cryos:
In future please try to separate your commits by purpose, i.e. removing paths as one commit, reverting changes as another. Commit subjects should be like an email subject, i.e. no more than 60 characters, no period at the end, short summary of the change. You can use paragraphs for more detail when needed. You don't need to change edit commits again, but please bear it in mind as you develop future branches. This is looking close, I would like @psavery to take a quick look.

examples/benchmark_stem_image.py
```diff
@@ -2,6 +2,7 @@
 from stempy import io, image
 import sys
 import time
+import numpy as np
```
cryos: I don't think this needs to be added anymore.

examples/stem_histogram.py
```diff
@@ -0,0 +1,85 @@
+import click
```
cryos: If you aren't using click I don't think you need to import the module.

examples/benchmark_stem_image.py
zhuokaizhao: Thanks, the line has been removed.

examples/stem_histogram.py
zhuokaizhao: Thanks, the line has been removed.

examples/stem_histogram.py
```diff
@@ -0,0 +1,84 @@
...
+    inner_radii = [0, 40]
+    outer_radii = [288, 288]
+
+    # file reader
+    reader = io.reader(files)
+
+    # generate histograms
+    numBins = 100
+    print('Generating histograms for input data')
+    allBins, allFreqs = image.create_stem_histogram(numBins, reader, inner_radii, outer_radii,
```
psavery: We use `camel_case` for variables in python, by the way.

examples/stem_histogram.py
```diff
@@ -0,0 +1,84 @@
...
+
+    stem_image_data = stem_image_data.reshape((160, 160))
+    stem_image_data = np.interp(stem_image_data, [min, max], [0, 256])
+    stem_image_data = stem_image_data.astype(np.uint8)
+    img = Image.fromarray(stem_image_data)
+    img.save(name)
+
+def main(argv):
+    print_help()
+    dataDir = argv[0] if len(argv) > 0
```
psavery: I get a syntax error here if I try to run it.

stempy/image.h
```diff
@@ -64,6 +64,11 @@ namespace stempy {
   template <typename InputIt>
   RadialSum<uint64_t> radialSum(InputIt first, InputIt last, int scanWidth = 0, int scanHeight = 0,
         int centerX = -1, int centerY = -1);
-}
+
+  // bins for histogram
+  std::vector<double> getContainer(const STEMImage& inImage, const int numBins);
+  // histogram and bins for the input STEM image
+  std::vector<int> createSTEMHistogram(const STEMImage& inImage, const int numBins, const std::vector<double> bins);
```
psavery: Let's run a `git-clang-format master` one time before merging to fix the formatting.

examples/stem_histogram.py
```diff
@@ -0,0 +1,84 @@
...
+    outDir = argv[1] if len(argv) > 1
+
+    # create output directory if it does not exit
+    if not os.path.exists(outDir):
+        os.mkdir(outDir)
+        print('Output directory', outDir , 'created')
+    else:    
+        print('Outout directory', outDir , 'already exists')
```
psavery: `Output`, not `Outout`

examples/stem_histogram.py
```diff
@@ -0,0 +1,84 @@
...
+def main(argv):
+    print_help()
+    dataDir = argv[0] if len(argv) > 0
+    outDir = argv[1] if len(argv) > 1
+
+    # create output directory if it does not exit
+    if not os.path.exists(outDir):
+        os.mkdir(outDir)
+        print('Output directory', outDir , 'created')
+    else:
```
psavery: There is some trailing white space in the file as well. I'd recommend getting your editor to automatically delete trailing white space upon saving the file.

examples/stem_histogram.py
```diff
@@ -0,0 +1,84 @@
...
+        # save to local
+        suffix = str(inner_radii[i]) + '_' + str(outer_radii[i]) + '.png'
+        plt.savefig(outDir + '/histogram_' + suffix)
+        
+        plt.show()
+
+
+if __name__ == "__main__":
+    main(sys.argv[1:])
+
```
psavery: A blank newline is here as well.

psavery:
Other than some formatting issues and a syntax error in the python sample, I think this is looking good! I think the histograms that it produces look good, and it doesn't look like other examples are affected by the changes anymore.

@cjh1 might want a jupyter notebook example as well (such as [this one](https://github.com/OpenChemistry/stempy/blob/master/examples/stem_image.ipynb)), but I think that can be done in another PR.

examples/stem_histogram.py
zhuokaizhao: Sorry for the typo, has been fixed.

examples/stem_histogram.py
zhuokaizhao: Great catch, I have removed it.

examples/stem_histogram.py
zhuokaizhao: I forgot to set this on with my intern laptop, sorry about that, just turned it on.

zhuokaizhao:
I have fixed the syntax error in the python sample, and deleted extra new line and trailing white spaces as spotted by @psavery . Thank you everyone.

stempy/image.h
zhuokaizhao: I believe that I've run `clang-format` on `image.h` before, also I hesitate to do it again because last time it changed lots of formats with the current old code and I had to manually change it back to what it was.

cryos: Let's just plan on a future run of clang-format on source files if this is the only remaining issue. Here it is clear that your line length is too high, I also suspect we should be preferring const ref for `bins` to avoid a copy.

psavery: Try using `git-clang-format` instead. It only applies the formatting to your own changes to the code. Make sure your master branch is up-to-date, and then run:

`git-clang-format master`

from inside your branch. You should get changes like the following:
```
diff --git a/stempy/image.cpp b/stempy/image.cpp
index 6e27107..06f92f4 100644
--- a/stempy/image.cpp
+++ b/stempy/image.cpp
@@ -17,13 +17,13 @@ using ArrayHandleView = vtkm::cont::ArrayHandleView<vtkm::cont::ArrayHandle<T>>;
 
 #endif
 
+#include <algorithm>
+#include <cmath>
 #include <future>
 #include <iostream>
 #include <memory>
 #include <numeric>
 #include <sstream>
-#include <cmath>
-#include <algorithm>
 
 using namespace std;
 
@@ -278,7 +278,8 @@ std::vector<double> getContainer(const STEMImage& inImage, const int numBins)
   int height = inImage.height;
   auto curData = inImage.data;
 
-  auto result = std::minmax_element(curData.get(), curData.get() + width * height);
+  auto result =
+    std::minmax_element(curData.get(), curData.get() + width * height);
   double min = *result.first;
   double max = *result.second;
 
@@ -296,7 +297,9 @@ std::vector<double> getContainer(const STEMImage& inImage, const int numBins)
 
 // function that computes histogram for all the STEM images
 // each histogram is a vector<int>
-std::vector<int> createSTEMHistogram(const STEMImage& inImage, const int numBins, const std::vector<double> bins)
+std::vector<int> createSTEMHistogram(const STEMImage& inImage,
+                                     const int numBins,
+                                     const std::vector<double> bins)
 {
   // initialize output
   std::vector<int> frequencies(numBins, 0);
@@ -317,7 +320,7 @@ std::vector<int> createSTEMHistogram(const STEMImage& inImage, const int numBins
     }
     // the max value is put in the last slot
     if (value == bins[numBins]) {
-        ++frequencies[numBins-1];
+      ++frequencies[numBins - 1];
     }
   }
 
diff --git a/stempy/image.h b/stempy/image.h
index 6ac184e..854a0eb 100644
--- a/stempy/image.h
+++ b/stempy/image.h
@@ -68,7 +68,9 @@ namespace stempy {
   // bins for histogram
   std::vector<double> getContainer(const STEMImage& inImage, const int numBins);
   // histogram and bins for the input STEM image
-  std::vector<int> createSTEMHistogram(const STEMImage& inImage, const int numBins, const std::vector<double> bins);
-} // namespace stempy
+  std::vector<int> createSTEMHistogram(const STEMImage& inImage,
+                                       const int numBins,
+                                       const std::vector<double> bins);
+  } // namespace stempy
 
 #endif
```

psavery:
Well done @zhuokaizhao! Thank you for your hard work!

PR #66, opened 2019-06-27
Add click as a requirement in requirements-dev.txt

psavery:
We use click in several scripts. Add it as a requirement.


cryos:
We use it in test scripts, but do we need it for the Python module itself? It would be good to keep the core module requirements really minimal, it is intended first and foremost as a Python module.

psavery:
Okay, yeah I think we can leave it out then. It's mostly just for test scripts. Maybe we can add the extra dev requirements in a `requirements-dev.txt` file?

cryos:
That could be the way to go, I think it is important to keep the main package as lean as reasonable.

psavery:
Done.

PR #67, opened 2019-06-28
Added contrast equalization (new) and piece-wise histogram generation support

zhuokaizhao:
Since most adaptive contrast equalization algorithms require piece-wise histograms of a single image, this feature has been added to the current histogram function.

I've also made some minor changes in Python example where variables are not in `camel_case` format.

Sample results with 5 histograms per STEM image look like:
histogram_0_40_288
histogram_1_40_288
histogram_2_40_288
histogram_3_40_288
histogram_4_40_288

Please feel free to let me know if there were any issues, thank you for reviewing.

Zhuokai


examples/stem_histogram.py
```diff
@@ -26,57 +26,69 @@ def save_img(stem_image_data, name):
...
-        plt.xticks(x[::20], bins[::20])
-        plt.title('Histogram of STEM image with inner radius = '
-                    + str(inner_radii[i]) + ', outer radius = ' + str(outer_radii[i]))
-        plt.xlabel('Value')
-        plt.ylabel('Frequency')
+        x = np.arange(num_bins+1)
+        for j in range(num_hist):
+            # init figure
+            fig = plt.figure(j, figsize=(16, 8))
+            myHist = fig.add_subplot(111)
```
psavery: You might want to my `myHist` here `camel_case` as well.

stempy/image.cpp
```diff
@@ -295,36 +295,59 @@ std::vector<double> getContainer(const STEMImage& inImage, const int numBins)
...
 
-// function that computes histogram for all the STEM images
+// function that computes numHist histograms for input STEM image
 // each histogram is a vector<int>
-std::vector<int> createSTEMHistogram(const STEMImage& inImage,
-                                     const int numBins,
-                                     const std::vector<double> bins)
+std::vector<std::vector<int>> createSTEMHistogram(
+  const STEMImage& inImage, const int numHist, const int numBins,
+  const std::vector<double> bins)
```
psavery: We probably should have done this in the original PR, but I think this should be `const std::vector<double>&` instead of `const std::vector<double>` to prevent an extra copy being made.

examples/stem_histogram.py
zhuokaizhao: Oh good catch, thanks!

stempy/image.cpp
zhuokaizhao: Sure, I have made the change, thanks!

psavery:
Feel free to make these changes to the jupyter notebook as well.

zhuokaizhao:
> Feel free to make these changes to the jupyter notebook as well.

Thank, I've made corresponding changes to the jupyter notebook. I probably did it in a wrong way though. I copied and pasted the current notebook to a newly-created file in my branch, which caused the conflict?

psavery:
Yeah, I think you can do one of two things here:

1. Revert that commit, merge master into this branch, then make the same changes again in a new commit, then force push (be careful force pushing, as changes can be erased).
2. Merge master into your branch right now, fix the conflicts, and then push.

psavery:
Make sure your master branch is updated before merging by first switching to the master branch (`git checkout master`), then pull (`git pull origin master` - assuming origin is `https://github.com/openchemistry/stempy`), then go back (`git checkout histogram`).

stempy/image.cpp
```diff
@@ -295,36 +295,59 @@ std::vector<double> getContainer(const STEMImage& inImage, const int numBins)
...
-    auto value = curData[i];
-    // check which bin it belongs to
-    for (int j = 0; j < numBins; j++) {
-      if (value >= bins[j] && value < bins[j + 1]) {
-        ++frequencies[j];
+    totalData[i] = curData[i];
+  }
+
+  // length of individual data
+  int subLength = numData / numHist;
```
psavery: I was taking a quick look at what happens here if `(width * height) % numHist != 0`. I printed out some debug output (11 histograms for this example):
```
width * height is: 25600
numData is: 25603
numHist is: 11
subLength is: 2327
```

Correct me if I'm wrong, but it looks like this will not cover all of the data? `subLength * numHist` is `25597`, which is 3 less than `width * height`. So it looks to me that in the loops below, the last three values of the image will not be placed in a bin?

stempy/image.cpp
```diff
@@ -295,36 +295,59 @@ std::vector<double> getContainer(const STEMImage& inImage, const int numBins)
...
+    // histogram of current piece
+    std::vector<int> curFrequencies(numBins, 0);
+
+    // get the histrogram of current area
+    for (int j = 0; j < subLength; j++) {
+      auto value = totalData[i * subLength + j];
+      // check which bin it belongs to
+      for (int k = 0; k < numBins; k++) {
+        if (value >= bins[k] && value < bins[k + 1]) {
+          ++curFrequencies[k];
```
psavery: One additional way we can also speed this up, I think, is to `break` out of the loop once we find the bin the value belongs in. It should only belong in one bin, correct?

This could also be done in the `if` block below.

zhuokaizhao: Yes it should only belong to one bin, I have added the `break` inside the `for` loop. Thanks.

stempy/image.cpp
zhuokaizhao: Yeah it is indeed a bug. The line of code that updates `numData` should actually be
`numData += (numHist - numData % numHist);`
instead of
`numData += numData % numHist;`

Thank you very much for testing it out!

zhuokaizhao:
Hi everyone, I have completed the standard contrast equalization that utilizes histograms for STEM images. Before equalization, an example STEM image with `inner_radius = 40` and `outer_radius = 288` may look like:
img_40_288
with the associated histogram
histogram_40_288
After contrast equalization, the image looks like
equalized_img_40_288
with the associated histogram
histogram_equalized_40_288
It makes sense to me that it is working, but feel free to let me know if anything looks wrong to you. Thank you for reviewing!

stempy/image.cpp
psavery: That is good! But we do also want to make sure that all of the data points get placed in bins, right? If `subLength * numHist` is 25597, doesn't that mean that the last 3 values (`width * height == 25600`) do not get placed in any bins? Perhaps you should add something to make sure that all of the leftover values (which will be equal to `(width * height) % numHist`) get placed in bins as well.

zhuokaizhao: The way I solved the issue you mentioned was by appending zeros to the "left-over values". Continue with your `numHist=11` example, the last 3 values will be appended with 8 zeros and all together be used to generate the histogram. Adding zeros does not affect anything because for our data, the min value is always positive (actually super large, which can be double-checked from the lower bound of the generated histogram), so having some additional zeros would not change the correctness of the histogram.

psavery: Oh yes, I see that now. Sounds good!

cryos:
It would be good to see the paper you are using. maybe it is the data set but this doesn't help highlight the nanoparticles as I would have expected. Looking at some other examples I would expect to see the noise reduced, and the features enhanced. Are there tunable parameters that might help here? It predominantly seems to increase the brightness.

cryos:
At the topic level branch, I would like you to avoid merging branches in, we can do that right before merge if necessary. We generally prefer rebasing, but that isn't usually needed. You should run git-clang-format before committing changes, or worst case as a final commit. It adds more noise to history, and removes value from git blame.

cryos:
[This article](https://www.hindawi.com/journals/mpe/2016/1529782/) shows some of the approaches I have seen before, and illustrative examples of enhancements possible.

PR #68, opened 2019-07-04
Release the GIL when calling C++ functions

cjh1:
This give other Python threads a look in!


cjh1:
@alesgenova I think this is the solution to the problem we where running to yesterday.

PR #73, opened 2019-07-10
add hdf5 reader

wangzhezhe:
1 add hdf5 reader which load the image array and generate the Block
2 the information to generate Header in Block should be set manually
3 tested ok for big_stem_image.h5

### DIFF
--- a/python/image.cpp
+++ b/python/image.cpp
@@ -20,7 +20,7 @@ PYBIND11_MODULE(_image, m)
           sizeof(uint64_t),                                             /* Size of one scalar */
           py::format_descriptor<uint64_t>::format(),                    /* Python struct-style format descriptor */
           2,                                                            /* Number of dimensions */
-          { i.width, i.height },  /* Buffer dimensions */
+          { i.height, i.width },  /* Buffer dimensions */
           { sizeof(uint64_t) * i.width,                           /* Strides (in bytes) for each index */
             sizeof(uint64_t) });
     });
@@ -32,7 +32,7 @@ PYBIND11_MODULE(_image, m)
           sizeof(double),                                               /* Size of one scalar */
           py::format_descriptor<double>::format(),                    /* Python struct-style format descriptor */
           2,                                                            /* Number of dimensions */
-          { i.width, i.height },  /* Buffer dimensions */
+          { i.height, i.width},  /* Buffer dimensions */
           { sizeof(double) * i.width,                           /* Strides (in bytes) for each index */
             sizeof(double) });
     });
@@ -44,7 +44,7 @@ PYBIND11_MODULE(_image, m)
           sizeof(uint64_t),                                               /* Size of one scalar */
           py::format_descriptor<uint64_t>::format(),                    /* Python struct-style format descriptor */
           3,                                                            /* Number of dimensions */
-          { r.radii, r.width, r.height},  /* Buffer dimensions */
+          { r.radii, r.height, r.width},  /* Buffer dimensions */
           { sizeof(uint64_t) * r.width * r.height,                           /* Strides (in bytes) for each index */
             sizeof(uint64_t) * r.width,
             sizeof(uint64_t) });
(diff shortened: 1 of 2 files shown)


stempy/reader.cpp
```diff
@@ -26,6 +26,24 @@ using std::vector;
...
+               uint32_t scanHeight, uint32_t startPosition)
+{
+    this->frameWidth = frameWidth;
+    this->frameHeight=frameHeight;
+    this->imagesInBlock = imageNumInBlock;
+    this->scanHeight=scanHeight;
+    this->scanWidth=scanWidth;
+    //only suitable for big_stem_image.hdf5
+    uint32_t offset1 = startPosition/(imageNumInBlock*blockNumInFile);
+    uint32_t offset2 = startPosition%(imageNumInBlock*blockNumInFile);
```
cjh1: This is specific to a particular dataset?

python/stempy/io/__init__.py
```diff
@@ -85,3 +86,6 @@ def save_stem_images(outputFile, images, names):
         stem_group = f.require_group('stem')
         dataset = stem_group.create_dataset('images', data=images)
         dataset.attrs['names'] = names
+
+def reader_h5(h5dataset, imagewidth , imageheight , scanwidth, scanheight, blocksize=32, blockNumInFile=25, totalImgNum=25600):
```
cjh1: The image width and height can be got from the h5 data set.

cjh1: We should also check for the presents of `Nx` and `Ny` attributes, these would hold the scan with and height.

CMakeLists.txt
```diff
@@ -81,10 +83,14 @@ endif()
...
 
+include_directories(stem
+  ${PYTHON_INCLUDE_PATH}
+  ${pybind11_INCLUDE_DIRS}
+)
 add_executable(stemcli stempy/main.cpp)
 
 target_link_libraries(stemcli
-  PRIVATE stem)
+  PRIVATE stem pybind11::embed)
```
cjh1: What is this needed for?

CMakeLists.txt
```diff
@@ -32,14 +32,16 @@ include_directories(
 find_package(Eigen3 REQUIRED)
 include_directories(${EIGEN3_INCLUDE_DIR})
 
+find_package(pybind11 REQUIRED)
```
cjh1: Aren't we already doing this is a subdirectory?

python/image.cpp
```diff
@@ -64,9 +65,9 @@ PYBIND11_MODULE(_image, m)
         py::call_guard<py::gil_scoped_release>());
   m.def("radial_sum", &radialSum<StreamReader::iterator>,
         py::call_guard<py::gil_scoped_release>());
-
   m.def("get_container", &getContainer,
         py::call_guard<py::gil_scoped_release>());
   m.def("create_stem_histogram", &createSTEMHistogram,
         py::call_guard<py::gil_scoped_release>());
+  m.def("create_stem_images", &createSTEMImages<H5Reader::iterator>);
```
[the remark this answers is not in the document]
wangzhezhe: This template is initialized by H5Reader::iterator, the previous one is initialized by StreamReader::iterator.

python/stempy/image/__init__.py
```diff
@@ -1,6 +1,19 @@
 from stempy import _image
 import numpy as np
 
+def create_stem_images_pyhdf5(h5reader, inner_radii,
```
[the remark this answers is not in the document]
wangzhezhe: I could update this part to use the original create_stem_images

stempy/reader.cpp
wangzhezhe: yeah, I only tested for big_stem_image, since the big_stem_image come from the smallscan data, and smallscan data has the header in dataset, here we set the header manually to avoid the modification for the original code.

python/stempy/io/__init__.py
wangzhezhe: I could update this part further.

CMakeLists.txt
wangzhezhe: Since the image.app include the reader_h5.h and reader_h5.h require the pybind so it is added here. Or another way is to add an option to decide if the reader_h5.h needed to be included in source code.

PR #78, opened 2019-07-16
add hdf5 reader

wangzhezhe:
1 add hdf5 reader which load the image array and generate the Block
2 the information to generate Header in Block should be set manually
3 tested ok for big_stem_image.h5

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,8 @@ include_directories(
 find_package(Eigen3 REQUIRED)
 include_directories(${EIGEN3_INCLUDE_DIR})
 
+find_package(pybind11 REQUIRED)
+
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 
 
@@ -39,7 +41,7 @@ set(_vtkm_src
   stempy/image.cpp
   stempy/electron.cpp
 )
-set(_src stempy/reader.cpp
+set(_src stempy/reader.cpp python/reader_h5.cpp
   stempy/mask.cpp
   stempy/electronthresholds.cpp)
 
@@ -81,10 +83,14 @@ endif()
 target_link_libraries(stem
   PRIVATE ${_libs})
 
+include_directories(stem
+  ${PYTHON_INCLUDE_PATH}
+  ${pybind11_INCLUDE_DIRS}
+)
 add_executable(stemcli stempy/main.cpp)
 
 target_link_libraries(stemcli
-  PRIVATE stem)
+  PRIVATE stem pybind11::embed)
 
 set_target_properties(stem
   PROPERTIES
@@ -99,3 +105,4 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/config.h.in"
 
 add_subdirectory(${PROJECT_SOURCE_DIR}/thirdparty/pybind11)
 add_subdirectory(python)
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,10 +32,21 @@ include_directories(
 find_package(Eigen3 REQUIRED)
 include_directories(${EIGEN3_INCLUDE_DIR})
 
-find_package(pybind11 REQUIRED)
+#find_package(pybind11 REQUIRED)
+add_subdirectory(${PROJECT_SOURCE_DIR}/thirdparty/pybind11)
+find_package(PythonInterp 3 REQUIRED)
+find_package(PythonLibs 3 REQUIRED)
+set(PYBIND11_PYTHON_VERSION "3" CACHE STRING "")
+set(PYBIND11_CPP_STANDARD "-std=c++14" CACHE STRING "")
+
+include_directories(SYSTEM
+  ${PYTHON_INCLUDE_PATH}
+  ${PYBIND11_INCLUDE_DIR}
+)
+
+message(STATUS "include path for pybind11 is: " ${PYBIND11_INCLUDE_DIR})
 
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
-
 
 set(_vtkm_src
   stempy/image.cpp
@@ -83,10 +94,6 @@ endif()
 target_link_libraries(stem
   PRIVATE ${_libs})
 
-include_directories(stem
-  ${PYTHON_INCLUDE_PATH}
-  ${pybind11_INCLUDE_DIRS}
-)
 add_executable(stemcli stempy/main.cpp)
 
 target_link_libraries(stemcli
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -32,14 +32,16 @@ include_directories(
 find_package(Eigen3 REQUIRED)
 include_directories(${EIGEN3_INCLUDE_DIR})
 
+find_package(pybind11 REQUIRED)
+
```
cjh1: As we are doing a find here, we should remove the extra call that is being done in the `/python/CMakeLists.txt`

examples/hdf5Reader/reader_hdf5.py
```diff
@@ -0,0 +1,29 @@
...
+    img.save(name)
+
+# init reader
+filename="./big_stem_image.h5"
+h5file= h5py.File(filename, 'r')
+
+inner_radii = [0, 40]
+outer_radii = [288, 288]
+
+imgs = image.create_stem_images(h5file, inner_radii, outer_radii, width=160, height=160)
```
cjh1: Nice, this look much cleaner.

python/reader_h5.cpp
```diff
@@ -0,0 +1,65 @@
...
+    m_blockNumInFile(blockNumInFile), m_totalImageNum(totalImageNum),
+    m_imageNumbers(imageNumbers)
+{}
+
+Block H5Reader::read()
+{
+
+  // get to the end of the file, return empty Block
+  if (m_currIndex >= m_totalImageNum) {
+    std::cout << "data loading finish with index " << m_currIndex << std::endl;
```
cjh1: We should remove this debug.

python/reader_h5.cpp
```diff
@@ -0,0 +1,65 @@
...
+  py::buffer_info buf = pyarray.request();
+  uint32_t arraySize = pyarray.size();
+
+  uint16_t* ptr = (uint16_t*)buf.ptr;
+
+  Header h = Header(m_imageWidth, m_imageHeight, m_imageNumInBlock,
+                    m_blockNumInFile, m_scanWidth, m_scanHeight, m_currIndex, m_imageNumbers);
+  Block b = Block(h);
+
+  std::copy(ptr, ptr + arraySize, b.data.get());
```
cjh1: Why are we doing this copy?

python/reader_h5.h
```diff
@@ -0,0 +1,98 @@
...
+
+    iterator(H5Reader* h5reader) : m_H5Reader(h5reader)
+    {
+      if(h5reader==nullptr){
+        return;
+      }
+      // read data at first time
+      m_block = m_H5Reader->read();
+      if (!m_block.data) {
+        std::cout<<"m_H5Reader is null" << std::endl;
```
cjh1: debug?

python/stempy/image/__init__.py
```diff
@@ -1,9 +1,40 @@
 from stempy import _image
+from stempy import _io
 import numpy as np
-
-def create_stem_images(reader, inner_radii,
+import h5py
+
+
+def get_hdf5_reader(h5file):
```
cjh1: Can we add a leading underscore to indicate that this is an internal function

python/stempy/image/__init__.py
```diff
@@ -1,9 +1,40 @@
...
+    # the initialization is at the io.cpp
+    dset_frame=h5file['frames']
+    dset_frame_shape=dset_frame.shape
+    totalImgNum=dset_frame_shape[0]
+    imagewidth=dset_frame_shape[1]
+    imageheight=dset_frame_shape[2]
+    
+    dset_stem_shape=h5file['stem/images'].shape
+    scanwidth=dset_stem_shape[1]
+    scanheight=dset_stem_shape[2]
```
cjh1: `stem/images` is unlikely to be present if we are trying to create a STEM image. We need to look for attributes `Nx` and `Ny`, that we can use to save the scan dimension in.

python/stempy/image/__init__.py
```diff
@@ -1,9 +1,40 @@
...
+    
+    # the number here are same with the smallscan data
+    blocksize=32
+    blockNumInFile=25
```
cjh1: If these are constants we should move them to the top of the file and use uppercase.

python/stempy/image/__init__.py
```diff
@@ -1,9 +1,40 @@
...
+    image_numbers = []
+    
+    h5reader = _io._h5reader(dset_frame, image_numbers, imagewidth, imageheight, scanwidth, scanheight, blocksize, blockNumInFile, totalImgNum)
+    return h5reader
+
+def create_stem_images(input, inner_radii,
                        outer_radii, width=0, height=0,
                        center_x=-1, center_y=-1):
+    # check if the input is the hdf5 dataset
+    if(isinstance(input, h5py._hl.files.File)):
```
cjh1: I think we can broaden this check to `hasattr(input, '__len__') and hasattr(input, '__getitem__')`, then is should also work with regular numpy arrays. We should then remain some the of the classes so they don't explicitly refer to HD5, for example `H5Reader` would become `SequenceReader`.

cjh1:
Looks like you need merge in master as well.

cjh1:
@alesgenova This is the HDF5 PR

alesgenova:
@wangzhezhe 
I'm trying to generate a stem image from a hdf5 file, but I'm running into issues. Not sure if the problem is in my code or in stempy.

Here is an example:
```python
import h5py
from stempy import io, image

filename = '/home/alessandro/Downloads/stem_image.h5'

h5file= h5py.File(filename, 'r')

inner_radii = (0,)
outer_radii = (40,)
imgs = image.create_stem_images(h5file, inner_radii, outer_radii, width=160, height=160)
```

It produces this debugging output when run
```
hdf5 reader is used
debug constructor
986
998
839
1077
902
1012
979
1027
1094
970
construct pyblock with index 0and 32
debug getData
pyblock get data is called 
pyblock ok to return shared pointer 
986
998
839
1077
902
1012
979
1027
1094
970
return block
m_H5Reader is null
```

And it crashes with the following trace
```
Traceback (most recent call last):
  File "test.py", line 11, in <module>
    imgs = image.create_stem_images(h5file, inner_radii, outer_radii, width=160, height=160)
  File "/home/alessandro/kitware/stempy-build/lib/stempy/image/__init__.py", line 40, in create_stem_images
    center_x, center_y)
ValueError: No blocks to read!
```

wangzhezhe:
Hello, @alesgenova  This branch is on progress and there are some bugs need to be fixed when we update the block into more efficient one. Please check this branch https://github.com/wangzhezhe/stempy/tree/feature/add_hdf5_reader_withoriginalblock , thanks!

PR #80, opened 2019-07-18
Fix overflow/wrap in processing sparse datasets

### LINKED ISSUE #77: create_stem_image_sparse
```create_stem_image_sparse``` does not seem to be working as expected. Half of the image is not the correct intensity compared to a non-counted STEM image.
image

Im not sure if this is due to counting or the STEM image creation.

I copied the notebook to ...../ncemhub/cjh1 on Cori.

### PR DESCRIPTION
**Pull request by cjh1**
This should resolve #77, I have pushed the images to cori.

### DIFF
--- a/stempy/image.cpp
+++ b/stempy/image.cpp
@@ -337,8 +337,9 @@ vector<uint16_t> expandSparsifiedData(const vector<vector<uint32_t>>& data,
 {
   vector<uint16_t> ret(data.size() * numPixels, 0);
   for (size_t i = 0; i < data.size(); ++i) {
-    for (auto pos : data[i])
+    for (auto pos : data[i]) {
       ret[i * numPixels + pos] = 1;
+    }
   }
 
   return ret;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: stempy/image.cpp [hunk 1]
```diff
@@ -142,14 +142,19 @@ void _runCalculateSTEMValues(const uint16_t data[],
...
 #endif
   for (unsigned i = 0; i < imageNumbers.size(); ++i) {
+    // We need to ensure we are using uint64_t to prevent overflow.
+    auto offset = static_cast<uint64_t>(i) * numberOfPixels;
 #ifdef VTKm
     // Use view to the array already transfered
-    auto view = vtkm::cont::make_ArrayHandleView(dataHandle, i * numberOfPixels,
+    // Note: We need to ensure the offset is pass as a uint64 to prevent
+    // overflow.
+    auto view = vtkm::cont::make_ArrayHandleView(dataHandle, offset,
```
psavery: It looks like the second argument to the `make_ArrayHandleView` function is a `vtkm::Id`, and this is defined as an `int64_t` as long as the cmake option `VTKm_USE_64BIT_IDS` is true (it is ON by default).

I hope `int64_t` is big enough (since `offset` is a `uint64_t`). As long as it is big enough, I think this should be good.

Note that if `VTKm_USE_64BIT_IDS` is off, `vtkm::Id` is `int32_t` instead.

cjh1: Yes, we should be good.

PR #81, opened 2019-07-29
add python reader for hdf5

wangzhezhe:
1 add python reader to load the array from the python especially for hdf5 in python
2 update the CMakeLists for checking the python interpreter

### DIFF
--- a/examples/hdf5Reader/reader_hdf5.py
+++ b/examples/hdf5Reader/reader_hdf5.py
@@ -1,0 +1,31 @@
+import h5py
+import sys
+from stempy import io, image
+import numpy as np
+from PIL import Image
+
+def save_img(stem_image_data, name):
+    min = np.min(stem_image_data)
+    max = np.max(stem_image_data)
+
+    stem_image_data = stem_image_data.reshape((160, 160))
+    stem_image_data = np.interp(stem_image_data, [min, max], [0, 256])
+    stem_image_data = stem_image_data.astype(np.uint8)
+    img = Image.fromarray(stem_image_data)
+    img.save(name)
+
+if(len(sys.argv)!=2):
+    print ("python3 reader_hdf5.py <file name>")
+    exit(0)
+
+filename=sys.argv[1]
+h5file= h5py.File(filename, 'r')
+
+inner_radii = [0, 40]
+outer_radii = [288, 288]
+
+imgs = image.create_stem_images(h5file, inner_radii, outer_radii, width=160, height=160)
+
+for i, img in enumerate(imgs):
+    suffix = str(inner_radii[i]) + '_' + str(outer_radii[i]) + '.png'
+    save_img(img, 'img_' + suffix)
--- a/python/pyreader.cpp
+++ b/python/pyreader.cpp
@@ -1,0 +1,70 @@
+#include "pyreader.h"
+#include <stempy/reader.h>
+
+namespace stempy {
+
+PyBlock::PyBlock(py::array_t<uint16_t> pyarray) : m_array(pyarray)
+{
+  this->data = DataHolder();
+  this->data.innerdata = m_array.data();
+}
+
+PyReader::PyReader(py::object pyDataSet, std::vector<uint32_t>& imageNumbers,
+                   uint32_t scanWidth, uint32_t scanHeight, uint32_t blockSize,
+                   uint32_t totalImageNum)
+  : m_pydataset(pyDataSet), m_scanWidth(scanWidth), m_scanHeight(scanHeight),
+    m_imageNumInBlock(blockSize), m_totalImageNum(totalImageNum),
+    m_imageNumbers(imageNumbers)
+{}
+
+PyBlock PyReader::read()
+{
+  // Acquire GIL before calling Python code
+  py::gil_scoped_acquire acquire;
+
+  // get to the end of the file, return empty Block
+  if (m_currIndex >= m_totalImageNum) {
+    return PyBlock();
+  }
+
+  // get the data pointer from the py object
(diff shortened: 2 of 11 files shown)


stempy/reader.cpp
```diff
@@ -23,6 +23,30 @@ using std::vector;
...
+    this->frameWidth = frameWidth;
+    this->frameHeight=frameHeight;
+    this->imagesInBlock = imageNumInBlock;
+    this->scanHeight=scanHeight;
+    this->scanWidth=scanWidth;
+    //if there is no imageNumbers in original data set, reconstruct it
+    if(imageNumbers.size()!=0){
+      this->imageNumbers=imageNumbers;
+    }else{
+      //this reconstruction is only tested ok for big_stem_image.hdf5
```
cjh1: I still don't understand this comment. Are you saying that this block is specific to a particular dataset?

python/stempy/image/__init__.py
```diff
@@ -1,9 +1,38 @@
...
+    dset_frame_shape=dset_frame.shape
+    totalImgNum=dset_frame_shape[0]
+    
+    dset_stem_shape=h5file['stem/images'].shape
+    scanwidth=dset_stem_shape[1]
+    scanheight=dset_stem_shape[2]
+    
+    # the number here are same with the smallscan data
+    blocksize=32
+    blockNumInFile=25
```
cjh1: Is number of blocks in file specific this dataset?

python/stempy/image/__init__.py
```diff
@@ -1,9 +1,38 @@
...
+    h5reader = _io._pyreader(dset_frame, image_numbers, scanwidth, scanheight, blocksize, blockNumInFile, totalImgNum)
+    return h5reader
+
+def create_stem_images(input, inner_radii,
                        outer_radii, width=0, height=0,
                        center_x=-1, center_y=-1):
+    # check if the input is the hdf5 dataset
+    if(isinstance(input, h5py._hl.files.File)):
+        reader = get_hdf5_reader(input)
+        print("python reader is used")
```
cjh1: Debug?

wangzhezhe: I could delete this info

python/stempy/image/__init__.py
wangzhezhe: yeah, since the small scan organize the data set like this, if the number is changed, there might be some problems. since the original image.cpp is programmed based on original smallscan data, I tried blocksize=16 blockNumInFile=50 , the outputted image is not correct.

examples/hdf5Reader/reader_hdf5.py
```diff
@@ -0,0 +1,35 @@
...
+
+    stem_image_data = stem_image_data.reshape((160, 160))
+    stem_image_data = np.interp(stem_image_data, [min, max], [0, 256])
+    stem_image_data = stem_image_data.astype(np.uint8)
+    img = Image.fromarray(stem_image_data)
+    img.save(name)
+
+# init reader
+#filename="/home/zhe/stempyData/big_stem_image.h5"
+#filename="/home/zhe/stempyData/big_stem_image_scanPositions.h5"
```
cjh1: Lets remove these

python/pyreader.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace stempy {
+
+PyBlock::PyBlock(py::array_t<uint16_t> pyarray) : m_array(pyarray)
+{
+  this->data = DataHolder();
+  this->data.innerdata = m_array.data();
+}
+
+PYReader::PYReader(py::object pyDataSet, std::vector<uint32_t>& imageNumbers,
```
cjh1: This should be `PyReader`

python/pyreader.cpp
```diff
@@ -0,0 +1,71 @@
...
+{}
+
+PyBlock PYReader::read()
+{
+  // Acquire GIL before calling Python code
+  py::gil_scoped_acquire acquire;
+
+  // get to the end of the file, return empty Block
+  if (m_currIndex >= m_totalImageNum) {
+    std::cout << "data loading finish with index " << m_currIndex << std::endl;
```
cjh1: We still have some debug statements.

stempy/reader.cpp
```diff
@@ -23,6 +23,25 @@ using std::vector;
...
+    this->imagesInBlock = imageNumInBlock;
+    this->scanHeight=scanHeight;
+    this->scanWidth=scanWidth;
+    //if there is no imageNumbers in original data set, reconstruct it
+    if(imageNumbers.size()!=0){
+      this->imageNumbers=imageNumbers;
+
+    }else{
+      //assume the image is consecutive if there is no image numbers included in the data source explicitly
+       throw std::runtime_error("the image_number is supposed to be provided by hdf5 file");
```
cjh1: I don't think we should be thrown an exception here? It probably wants to be from where we construct the Head?

cjh1:
@alesgenova This is ready for you to test. @wangzhezhe Can point you to a new HDF5 file to test with.

wangzhezhe:
the new file is here on ulex: /home/zhe/stempyData/big_stem_image_scanPositions.h5 @cjh1 @alesgenova

python/pyreader.cpp
wangzhezhe: updated at the latest commit.

examples/hdf5Reader/reader_hdf5.py
wangzhezhe: updated at the latest commit.

python/pyreader.cpp
wangzhezhe: updated at the latest commit.

Issue #76, opened 2019-07-15
electron_count improvements


ercius:
1. The value for the background_threshold_n_sigma and x_ray_ variables in:
```stempy.image.electron_count```
only accept integers. This should accept float values as well. 

2. Please provide options for verbose output. It would be very nice to know the parameters of the gaussian fit used to determine the sigma value. center, sigma, and maximum.

3. width/height and frame_width/frame_height should be determined from the reader which loaded the data.

cjh1:
For 3. we do determine the width and height from the reader in the `electron_count(...)`, those parameters are optional. I assume we are talking about `create_stem_images_sparse(...)` in which case we require those parameters as we are passing in the sparse dataset with doesn't have this information. We can look at attaching it in some way to make those parameters options.

ercius:
Yes. That was what I wanted. The sparse data output should retain the
metadata.

cjh1:
Note for future implementer. I think for 3. we should change the return type of [electronCount](https://github.com/OpenChemistry/stempy/blob/master/stempy/electron.h#L9) to be a small structure containing the `vector`, we can then add the scan and frame dimensions to the structure. We will then need add the appropriate buffer protocol definition so the the new structure can still be used as a numpy array ....

cjh1:
@ercius I have just updated the images on cori that resolves 1 and 3. Note that `electron_count(...)` now returns an object that has a property called `data` which is the numpy array, as well as properties for the various dimensions. This object can be passed directly to `create_stem_images_sparse(...)`, removing the need to pass in all the dimensions.

ercius:
@cjh1 That is great! It makes the processing much easier. Thanks a lot.

PR #85, opened 2019-08-01
Add verbose option to electron threshold function

psavery:
The verbose option causes the electron threshold function
to print out to the console a variety of variables. These
are the current variables that are printed out:

```
numberOfSamples
min
max
mean
variance
stdDev
numberOfBins
xRayThresholdNSigma
backgroundThresholdNSigma
xrayThreshold
backgroundThreshold
```

Fixes: #76

### DIFF
--- a/python/image.cpp
+++ b/python/image.cpp
@@ -51,6 +51,24 @@ PYBIND11_MODULE(_image, m)
             sizeof(uint64_t) });
     });
 
+  py::class_<CalculateThresholdsResults>(m, "_calculate_thresholds_results",
+                                         py::buffer_protocol())
+    .def_readonly("background_threshold",
+                  &CalculateThresholdsResults::backgroundThreshold)
+    .def_readonly("xray_threshold", &CalculateThresholdsResults::xRayThreshold)
+    .def_readonly("number_of_samples",
+                  &CalculateThresholdsResults::numberOfSamples)
+    .def_readonly("min_sample", &CalculateThresholdsResults::minSample)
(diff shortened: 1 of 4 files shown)


psavery:
@ercius This works towards fixing number 2 in #76. But I have two questions:

1. I see that you asked for printing of "center, sigma, and maximum". I looked through our variables and the `lsq` library, and I don't see "center" as a variable anywhere. Perhaps it is something we need to calculate? Or, if the above list of variables is good enough, we can leave it.

We have two sigma's printing out (as can be seen above). Let me know if those are good for sigma.

2. Are you using a jupyter notebook for these? Unfortunately, `cout` in C++ is printing directly to the console, and it is not printing in the jupyter notebook. If you aren't using the notebook, then it might not matter right now.

ercius:
@psavery 

1. Center is the same as mean. So, the verbose output above looks good to me.

Sigma's look good. I think that is all we need.
2. I am currently using a Jupyter notebook for processing. Its the most convenient way to interact with the data at the moment. Im not sure how that will change in the future, but I think it would be great to have this available in the notebook. Can it be returned from the function if verbose is selected? Im open to any ideas, or leaving it as-is if its too difficult to capture the output in the notebook.

psavery:
@ercius I modified it so that the C++ returns the statistics from the electron threshold function. The python will then print these values automatically if you pass `verbose=True` to the `image.electron_count` function. An example can be seen in the attached image.
Screenshot from 2019-08-05 12-47-49

ercius:
That's great. Looks good to me.

cjh1:
@ercius This images on cori now contain this change.

PR #88, opened 2019-08-05
add python reader for hdf5

wangzhezhe:
add python reader to support hdf5


cjh1:
@wangzhezhe I Fixed up a few warnings and the formatting. This is now working as expected. The performance at the moment see to be pretty bad, but we can address that at a later date.

wangzhezhe:
Thanks for your help! @cjh1

PR #89, opened 2019-08-08
Add example of creating stem images in parallel

cjh1:


examples/create_stem_mpi.py
```diff
@@ -0,0 +1,81 @@
...
+    comm.Barrier()
+    start = MPI.Wtime()
+
+    files = get_files(files)
+
+    # Create local sum
+    reader = io.reader(files, version=io.FileVersion.VERSION3)
+
+    # Get the scan image size
+    block = reader.read()
```
psavery: By the way, if this ends up taking a lot of time, maybe in the future we can add a function to the `StreamReader` to just read a header and return it, rather than reading the whole block.

cjh1: Not a bad idea

PR #92, opened 2019-09-04
Make changes to use the master branch of vtkm

psavery:
"Invoke" was placed in "cont" rather than "worklet".

Also, "CellSetStructured" does not accept a string in the
constructor anymore. It may not be needed, just remove it.


psavery:
Might want to test this just a bit before merging.

psavery:
I've tested this out on our basic examples of creating a stem image and electron counting, and it seems to work okay.

PR #93, opened 2019-09-05
Add electron MPI example

psavery:
This is an example of performing the electron counting using MPI. It seems to work, and it provides a decent speed increase as well.

One concern we had was that the background thresholds and x-ray thresholds may not be as accurate, because each MPI process can currently only sample from its own files. However, for our test dataset, it does not appear to produce a noticeable difference.

Timings are interesting. On ulex, with `OMP_NUM_THREADS=1`, vtkm with openmp, and disk caching in use, these are the timing results:
```
Num MPI Processes: Seconds
1: 66.93
2: 66.80
4: 33.42
8: 19.14
```

* Note for our example, we have 8 files. Four of the files are 2.5 GB, and four are 100 MB. The lack of improvement from one MPI process to two is because one process is reading/processing all of the large files, and the other process is reading/processing all the small files.

### DIFF
--- a/examples/electron_mpi.py
+++ b/examples/electron_mpi.py
@@ -93,20 +93,6 @@ def main(files, dark_file, center, inner_radii, outer_radii, output_file,
         io.save_electron_counts(output_file, global_frame_events, scan_width,
                                 scan_height, frame_width, frame_height)
 
-        # Save out the electron counted image
-        img = np.zeros((576, 576))
-
-        # Just sum the events for now
-        for frame in global_frame_events:
-            for pos in frame:
-                row = pos // 576
-                column = pos % 576
-                img[row][column] += 1
-
-        fig, ax = plt.subplots(figsize=(12, 12))
-        ax.matshow(img)
-        plt.savefig('electron.png', dpi=300)
-
         if generate_sparse:
             # Save out the sparse image
             width = electron_counted_data.scan_width
--- a/examples/electron_mpi.py
+++ b/examples/electron_mpi.py
@@ -25,9 +25,12 @@ def get_files(files):
               required=True)
 @click.option('-o', '--outer-radii', help='The outer radii (comma separated)',
               required=True)
+@click.option('-f', '--output-file', help='The output HDF5 file to write',
+              default='electron_counted_data.h5')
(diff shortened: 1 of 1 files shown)


psavery:
Interestingly, without disk caching, mpi using 8 process each with 1 OMP thread runs faster than one process that has 20 OMP threads (which took around 25 seconds).

However, if the disk caches are cleared, one process with 20 OMP threads is faster (100.54 seconds vs. 154.98 seconds).

cjh1:
@psavery Can you add the example invocation that you where using for scan30?

examples/electron_mpi.py
```diff
@@ -74,6 +74,16 @@ def main(files, dark_file, center, inner_radii, outer_radii, generate_sparse):
...
+        # Write out the HDF5 file
+        scan_width = electron_counted_data.scan_width
+        scan_height = electron_counted_data.scan_height
+        frame_width = electron_counted_data.frame_width
+        frame_height = electron_counted_data.frame_height
+
+        output_file = 'electron_counted_data.h5'
+        io.save_electron_counts(output_file, global_frame_events, scan_width,
+                                scan_height, frame_width, frame_height)
+
```
psavery: @cjh1 I added a few lines of code here that saves out the electron counted data into an HDF5 file.

psavery:
@cjh1 Done.

cjh1:
We need to make this gracefully handle the situation where more tasks the files are allocated ( we should probably print a warning as well, as it not very useful ). At the moment if that happens, things are terminated as some of the readers complain they have know files to read ...

psavery:
@cjh1 I added a check that does as we discussed.

examples/electron_mpi.py
```diff
@@ -0,0 +1,179 @@
...
+        # Just sum the events for now
+        for frame in global_frame_events:
+            for pos in frame:
+                row = pos // 576
+                column = pos % 576
+                img[row][column] += 1
+
+        fig, ax = plt.subplots(figsize=(12, 12))
+        ax.matshow(img)
+        plt.savefig('electron.png', dpi=300)
```
cjh1: We should make this optional as well or just remove it.

psavery: Done.

PR #94, opened 2019-09-12
Add some missing reqs to requirements-dev

psavery:
These are required in some of the example files.


psavery:
@cryos We had actually created this file with the intention of putting requirements for the examples in it. The requirement that is already inside, `click`, is also only a requirement for the examples.

Maybe we should move it into the examples dir as `requirements.txt` if that would be better?

cjh1:
@psavery Yes, moving it to the examples directory makes more sense.

cryos:
Completely agree, examples are nice, and we want to avoid making people guess, but they shouldn't be confused with developer requirements which are often more than is needed to simply run the code. I hate encountering packages that attempt to install too much - keep our requirements as lean as we reasonably can.

psavery:
Sounds good, I moved that requirements file into the examples directory.

PR #95, opened 2019-09-24
Do not expand sparse data for sparse stem images

psavery:
When the sparse stem image functions were originally created, they
expanded the sparse data so that they could re-use functions that
are used by regular stem images. This, however, creates a very large
copy for some datasets.

This modification avoids that expansion and simplifies sparse stem
image creation. It should, in theory, be faster as well.

### DIFF
--- a/stempy/image.cpp
+++ b/stempy/image.cpp
@@ -343,8 +343,9 @@ void calculateSTEMValuesSparse(const vector<vector<uint32_t>>& data,
 {
   for (unsigned i = 0; i < data.size(); ++i) {
     uint64_t values = 0;
-    for (auto pos : data[i])
+    for (auto pos : data[i]) {
       values += mask[pos];
+    }
     image.data[i] = values;
   }
 }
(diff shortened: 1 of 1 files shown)


stempy/image.cpp
```diff
@@ -338,17 +338,15 @@ std::vector<int> createSTEMHistogram(const STEMImage& inImage,
...
 {
-  vector<uint16_t> ret(data.size() * numPixels, 0);
-  for (size_t i = 0; i < data.size(); ++i) {
-    for (auto pos : data[i]) {
-      ret[i * numPixels + pos] = 1;
-    }
+  for (unsigned i = 0; i < data.size(); ++i) {
+    uint64_t values = 0;
+    for (auto pos : data[i])
+      values += mask[pos];
```
cjh1: Lets use `{ }` for the for loop, even if there is one line at the moment. Is so easy to add another line without thinking and it not part of the loop!

psavery: Done!

PR #99, opened 2019-11-15
Extend the pipeline metadata with input/output types

alesgenova:


python/stempy/pipeline/__init__.py
```diff
@@ -2,6 +2,17 @@
...
+class PipelineIO:
+    Image = 'image'
+    Frame = 'frame'
+
+class PipelineAggregation:
+    Sum = 'sum'
+    Max = 'max'
+    Min = 'min'
+    Mean = 'mean'
+    Median = 'median'
```
cjh1: Lets make these upper case

alesgenova: Done!

PR #101, opened 2019-12-12
Update to support sector bases blocks

cjh1:


cjh1:
@cryos Could you take a quick look at his so I can get it merged?

CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ include_directories(
 )
 
 find_package(Eigen3 REQUIRED)
-include_directories(${EIGEN3_INCLUDE_DIR})
+include_directories(SYSTEM ${EIGEN3_INCLUDE_DIR})
```
cryos: Yes, this is really useful for isolating warnings from third party libraries (where we can't fix them) versus our code (which we should fix warnings in).

CMakeLists.txt
```diff
@@ -77,7 +84,7 @@ endif()
 set_property(TARGET stem PROPERTY POSITION_INDEPENDENT_CODE ON)
 
 target_link_libraries(stem
-  PRIVATE ${_libs})
+  PRIVATE "${_libs}")
```
cryos: Why the quotes here? Lists should expand as expected without them - is there an issue with HDF5 needing this?

stempy/reader.cpp
```diff
@@ -23,6 +26,11 @@ using std::vector;
 
 namespace stempy {
 
+const int SECTOR_WIDTH = 144;
```
cryos: It would be good to add a brief comment explaining why these are hard-coded for our future selves.

CMakeLists.txt
cjh1: Not sure why, I removed them and it work fine.

Issue #108, opened 2020-01-17
Clean up x, y vs row, col vs width, height in C++ code


cjh1:
While we are doing #106, I would like to do a similar pass through the C++ code. I can see at least one place where there may be a discrepancy:

https://github.com/OpenChemistry/stempy/blob/master/stempy/image.h#L53

cjh1:
I think we should use `std::pair` to represent dimensions and x, y coordinates ( I think then pybind11 will take care of the convertions for us). We can create a few type definitions for example:

`using Dimensions = std::pair<int, int>; // width, height`
`using Coordinates = std::pair<int, int>;

or alternatively we could create a new simple `struct`'s, but I'm not sure how we would get pybind11 to do the correct convertion from `tuple`.

x, y = width, height. Lets replace any uses of rows and columns.

alesgenova:
Unless I'm getting confused, I think the convention in numpy is the following though:

```python
shape = (n_rows, n_cols)  # i.e. (height, width)
```

cjh1:
@alesgenova This is true but I guess in general when talking about images or 2d graphic x is the width and y is the height ...

psavery:
@cjh1 For that example you mentioned [here](https://github.com/OpenChemistry/stempy/blob/master/stempy/image.h#L53), it looks like those two variables are mistakenly labeled `rows` and `columns` in the header file, when in the CPP file and in all of the use cases of it (in python included), they are treated as `width` and `height`. So it looks to me that it is just a labeling problem in the header, and that gets fixed in #109.

psavery:
However, I did end up switching [these two variables](https://github.com/OpenChemistry/stempy/pull/109/files#diff-5afbef7ae6ecfa409c949d37ebfd1e93R90) in #109, as they seemed out of order. We'll probably need to do some tests on some non-square data to confirm.

ercius:
@psavery @cjh1 I added an example data set which is non-square and small (11GB). See `ncemhub/4Dcamera/2020.01.13/data_scan137*.data`. You can count this in a few seconds (background_threshold_n_sigma = 4; background_threshold = 28.62) and create stem images very quickly too. If you get the scan width and height incorrect it's obvious to see the mistake in a stem image. The original data was acquire with 129 columns (x, width) and 128 rows (y, height) on the microscope. So, stempy should try to match this.

The correct ADF was created using this python command:
```python
ims = stim.create_stem_images_sparse(ee,(0,200),
[...]

Issue #107, opened 2020-01-16
Add doc strings to all public python functions


cjh1:
and setup readthedocs ...

cjh1:
I think we should use sphinx here for its ability to generate documentation from doc strings. I would like to use mkdocs, but it doesn't have great support for API documentation. 

Here is the getting started guide for sphink: https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html

cjh1:
I think we can still use MarkDown with sphinx.

ercius:
I use sphinx on ncempy. The auto api documentation generation is very
useful. It can be difficult to get up and running though and I think I
still am not doing it right.

psavery:
As a side note here, I used a [markdown to rst converter](https://pypi.org/project/m2r/) for one file we had already written in markdown, and it worked very well!

Issue #106, opened 2020-01-15
width vs height vs x vs y vs...


ercius:
For non-square scans Im getting confused as to which order to put the scan values in.

We should think about how to unify the inputs and outputs to all functions in terms of which axis is desired when variables like height and width are used as well as variable names with x and y.

Personally, I prefer to use tuples for these kinds of inputs and outputs. For example, if you have a 4D array of shape `[scany,scanx,ky,kx]` it can be easier to have a function input like:
```python
def my_func(data, scan_x_y, frame_x_y):
    ''' arbitrary function
    Inputs
        data : ndarray
                The data to input
        scan_x_y: tuple, 2-element
                The scan size which corresponds to the first and second data axes
        frame_x_y: tuple, 2-element
                The size of each detector frame corresponding to the third and fourth axes
    '''
```
Thus, the tuple order matches the order in the data. This gets away from width/height and x/y where you might be confused as to which is which. Is the (x, y) order like humans are used to or (y,x) order which is more like C-Ordering. Does width == x and height == y? Or the other way around

cjh1:
@ercius I agree this is an area of much confusion! Even writing this, I had to check the code! I think spending some time to clear this up is a good idea. Yes, currently x,y == width, height == columns, rows, if there is a more familiar convention we can switch to it?

cjh1:
As part of this adding doc strings to function would also help.

cryos:
Major thing would probably be more explicitly documenting. I would expect all pairs to be x, y and width, height. Using tuples seems like a reasonably approach, writing x = 10, y = 5 could help too, then you can be very explicit about them in Python at least.

cjh1:
@cryos your right using named parameters is another option, which prevent getting things out of order. I was trying to look for prior art in the numpy world, I guess the most basic example is the construction of an n-d array, where the dimensions are passed as a tuple.

cjh1:
Look like this is related #54 ( we have thought about this before :-) )

ercius:
Haha. I guess that was a good idea even back then!

cjh1:
Yep, I think we should go with the tuple approach.

cjh1:
I want to make sure we get this right, so thought some diagrams might help clarify things (@ercius I think the x and y for the parameters are in fact reversed, they don't match the order in the data? ) If we are looking at a scan or frame the coordinate it looks like the following.

(0,0)                    x
    + ------------------------------------ >
    |
    |
y  |
    |
    |
   +
[...]

Issue #54, opened 2019-05-27
Think about combining some parameters into tuples


cjh1:
Such as center_x, center_y and width and height ...

psavery:
What about a named tuple (and possibly a C++ struct) to access them by name?

cjh1:
or std::pair on the C++ side

PR #109, opened 2020-01-22
Use tuples and std::pair for dimensions and points

psavery:
Rather than having separate variables for things such as width and
height, combine them into one variable. In python, this is a tuple, and
in C++, this is a std::pair.

This combines scan width and scan height into "scan_dimensions" (python)
and "scanDimensions" (C++) variables. Frame width and frame height are
combined into "frame_dimensions" (python) and "frameDimensions" (C++).

Center x and Center y are combined into just a "center" variable.

Typically, dimensions.first is the width, and dimensions.second is the
height. But there is one place where the ordering was switched, because
it appeared to be wrong: maximalPointsParallel in electron.cpp. This
needs to be tested with a non-square dataset.

Fixes: #54
Fixes: #106
Fixes: #107
Fixes: #108

### DIFF
--- a/stempy/electron.cpp
+++ b/stempy/electron.cpp
@@ -15,6 +15,8 @@
 #include <sstream>
 #include <stdexcept>
 
+using stempy::Dimensions2D;
+
 #ifdef VTKm
 namespace {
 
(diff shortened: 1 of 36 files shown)


stempy/reader.h
```diff
@@ -13,25 +14,27 @@ class H5ReadWrite;
 
 namespace stempy {
 
+using Coordinates2D = std::pair<int, int>;
+using Dimensions2D = std::pair<uint32_t, uint32_t>;
```
cjh1: Lets add a comment here to explain the convention as in coordinate x,y and dimensions width, height ...

stempy/electron.cpp
```diff
@@ -123,14 +128,17 @@ inline uint16_t mod(uint16_t x, uint16_t y)
...
+  auto width = frameDimensions.first;
+  auto height = frameDimensions.second;
+
   std::vector<uint32_t> events;
   auto numberOfPixels = height * width;
   for (int i = 0; i < numberOfPixels; i++) {
-    auto row = i / width;
-    auto column = i % width;
+    int row = i / width;
+    int column = i % width;
```
cjh1: Why `int` here?

psavery: Similar to [here](https://github.com/OpenChemistry/stempy/pull/109/files/e54292bdde442e913199e200b21bfc2dd25cc4e4#diff-93e8605537e636aa2f094e4472e46b24R26), I made `row` an int so that `row - 1` (see below) won't do an unsigned integer underflow. I'm **think** `row` can be 0. `column` doesn't need to be an `int`, but I just did it for consistency...

psavery: @cjh1 and I discussed this, and we realized that we need to make sure the pixels on the borders don't try to check neighbors outside the image (like, currently, the pixels on the top border are trying to compare pixels above them). Once we change that, we will not need to make the row and column into `int` anymore either.

psavery: This is fixed.

stempy/reader.h
psavery: Fixed.

PR #113, opened 2020-01-30
Mock out pybind11 generate classes

cjh1:
We are doing this manually to avoid problems with nested classes.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -12,6 +12,8 @@
 
 import os
 import sys
+from unittest import mock
+
 sys.path.insert(0, os.path.abspath('../../python/stempy'))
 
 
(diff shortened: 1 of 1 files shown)


cjh1:
@psavery I think this resolves the issue you where running into. Let me know.

psavery:
@cjh1 Yes, this does appear to work!

It's a little unfortunate that we can't use sphinx's mock classes and their `autodoc_mock_imports` variable for this. But, when we have a class that inherits from their Mock classes, and we try to use that class for a function's default arguments, we run into attribute errors.

I was also thinking to mock `numpy` and `h5py` as well, so that RTD doesn't need to install them for every build.

psavery:
By the way, I had also figured out how to disable that constant expansion that we were seeing with default arguments before (like below):

image

Apparently, if you put a function signature as the first line in the function's docstring, sphinx will use that signature instead of generating one on it's own, like this:

```
def reader(path, version=FileVersion.VERSION1):
    """reader(path, version=FileVersion.VERSION1)
    ...
    """
```

This generates the function signature for the docs without the constant expansion. And this also applies to our case we are fixing here with `SectorReader.H5Format.Frame`.

cjh1:
@psavery Absolutely you can take over this PR.

PR #117, opened 2020-01-31
Add instruction for singularity

cjh1:


cjh1:
@ercius Here are the instructions for using singularity images, its a nice approach that you might want to try, which avoids having to build anything. Let me know if you have questions.

ercius:
@cjh1 Thanks! I'll try this next week. It will be great to avoid l avoid build issues when updating.

PR #121, opened 2020-02-10
Lock down vtk-m version in Dockerfile to 1.5

psavery:
It was previously building master, and updates to master could potentially
break the Dockerfile. Lock down the version so we can get a consistent
docker image.

Fixes: #120

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -51,9 +51,9 @@ RUN cd /build/mpi4py-3.0.1 && python3 setup.py build && python3 setup.py install
 
 # Build VTK-m
 RUN cd /source && \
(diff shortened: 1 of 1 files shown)


psavery:
@cjh1 I wonder if [this sha](https://github.com/OpenChemistry/stempy/blob/7f4df7fe9abcf77142ca5942a22bf19e9121b8cb/docker/Dockerfile#L30) will be changing anytime soon?

cjh1:
> @cjh1 I wonder if [this sha](https://github.com/OpenChemistry/stempy/blob/7f4df7fe9abcf77142ca5942a22bf19e9121b8cb/docker/Dockerfile#L30) will be changing anytime soon?

I would hope not

Issue #104, opened 2020-01-08
building stempy with sector reader


ercius:
I just wanted to leave a couple of quick comments about building stempy on an Ubuntu 18.04 machine.

I followed the script [here](https://github.com/OpenChemistry/stempy/blob/master/docker/Dockerfile) but had to make some modifications. These might not be all of the needed modifications as I was not keeping track as I solved issues.

1) I needed to switch the to the sector-reader branch to get io.FileVersion.VERSION4
2) I needed to manually clone pybind11 into the thirdparty directory
3) I needed to install (apt-get) libhdf5-dev for this branch
4) I had already installed python3 (3.6.9) from apt-get. So, I used that python rather than getting the version from the ftp on line 36. Not sure which way would be the best way to go.
 - Also, the step where you copy the *.so files needed to go to a different place in /home/username/.local/python3/site-packages/... 
5) I skipped VTKm. Im not sure if that part of the script is up to date.
6) I used pip3 to install mpi4py rather than the manual way. I could not get distutils working in the apt-get version of python3 from the Ubuntu repositories.

cjh1:
For step 2. git clone --recursive would do the trick ...

ercius:
> Also, the step where you copy the *.so files needed to go to a different place in /home/username/.local/python3/site-packages/...
The path is actually:
/home/username/.local/lib/python3.6/site-packages/stempy/.

PR #123, opened 2020-02-11
Update BUILDING.md

psavery:
This also adds an example of installing the stempy dependencies
on Ubuntu 18.04

Here are some updates:

* Added HDF5 as a dependency
* Use VTK-m version 1.5
* Added CMake 3.12+ example for Ubuntu 18.04
* Updated stempy lib soft link to more likely location
* Added clarification of `site-packages` location

Fixes: #104


psavery:
Here is the current order of the table of contents. Let me know if the order should be modified.

Screenshot from 2020-02-11 17-21-03

cjh1:
@psavery The only change I would make is I think stempy should lowercase, other than that it looks good.

psavery:
@cjh1 Done!

Issue #84, opened 2019-07-31
Look a providing unified overloaded create_stem_images(...) call


cjh1:
Rather than have `create_stem_images(...)` and `create_stem_image_sparse(...)` have a single implementation that can take either data type.

PR #124, opened 2020-02-12
Overload create stem image functions

psavery:
Rather than having separate functions for sparse STEM image creation
such as "create_stem_image_sparse", combine all of the sparse STEM
image functions (and the singular "create_stem_image" function) into
"create_stem_images".

"create_stem_images" determines the type of its input, and calls the
correct function signature based upon it. Now, a user can pass to
"create_stem_images" one of four options:

1. A stempy.io.reader
2. An open h5py file
3. ElectronCountedData (returned from electron counting) sparse data
4. A numpy array of sparse data

The "create_stem_images" function handles all of these inputs.

Additionally, on the C++ side, the "...Sparse()" functions were renamed
to overload "createSTEMImages".

This reduces the number of function names that we have to work with,
and it allows the code to automatically determine which functions to
call based upon the input.

This also adds and uses a deprecation package which automatically
generates deprecation information in the sphinx documentation for
deprecated functions.

Fixes: #84

### DIFF
--- a/python/stempy/image/__init__.py
+++ b/python/stempy/image/__init__.py
@@ -1,16 +1,21 @@
 from stempy import _image
 from stempy import _io
-import numpy as np
-import h5py
+from stempy.io import get_hdf5_reader, ReaderMixin
+
 from collections import namedtuple
-from stempy.io import get_hdf5_reader
+import deprecation
+import h5py
+import numpy as np
 
 def create_stem_images(input, inner_radii, outer_radii, scan_dimensions=(0, 0),
-                       center=(-1, -1)):
+                       center=(-1, -1), frame_dimensions=None, frame_offset=0):
     """Create a series of stem images from the input.
 
-    :param input: the file reader that has already opened the data.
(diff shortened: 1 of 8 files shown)


psavery:
To do: allow both raw and sparse data to be passed into `create_stem_images()` as numpy arrays.

python/stempy/image/__init__.py
```diff
@@ -26,68 +33,74 @@ def create_stem_images(input, inner_radii, outer_radii, scan_dimensions=(0, 0),
...
-        allBins.append(bins)
-        allFreqs.append(freq)
-
-    return allBins, allFreqs
-
-# This one exists for backward compatibility
+@deprecation.deprecated(
+    deprecated_in='1.0', removed_in='1.1',
+    details='Use stempy.image.create_stem_images() instead.')
 def create_stem_image(reader, inner_radius, outer_radius,
```
cjh1: I do wonder is we should just keep this one around ( forward the call on ). Thoughts?

Issue #129, opened 2020-02-17
Update README.md example


psavery:

ercius:
This can be closed.

PR #130, opened 2020-02-17
Update the example in README.md

psavery:
Instead of the previous example, show an example of creating STEM
images, and some example output.

Fixes: #129


cjh1:
@psavery It would be good to resolve this if it still relevant ...

ercius:
I think we are good with the readme for now.

cjh1:
Closing this for now.

Issue #126, opened 2020-02-13
PyReader requires "stem/images" for scan dimensions


psavery:
[Here](https://github.com/OpenChemistry/stempy/blob/7f4df7fe9abcf77142ca5942a22bf19e9121b8cb/python/stempy/io/__init__.py#L60-L61) in `get_hdf5_reader()`, the scan dimensions are determined by STEM images in the HDF5 file. However, it is not guaranteed that there will be STEM images in the HDF5 file (for instance, there may just be raw data).

We should come up with a way to either maybe save the scan dimensions into the HDF5 file, or have the user pass the scan dimensions in and use those for `get_hdf5_reader()`.

This issue comes from [this comment](https://github.com/OpenChemistry/stempy/issues/111#issuecomment-584650279) in #111.

cjh1:
@psavery I though we where saving the scan dimensions are attributes in the HDF5 file?

psavery:
@cjh1 I do see [from the documentation](https://stempy.readthedocs.io/en/latest/StempyH5Format.html) that it is being saved in `electron_events/scan_positions` as the `Nx` and `Ny` attributes. So we could potentially extract it from there as well, but it would only be there if the HDF5 file contained electron counted information.

It's also present if the raw data at `/frames` if it is in the square format (the 4-D format, where the scan dimensions will be the first and second dimensions).

I don't think it's saved anywhere else, though. It wouldn't be present if the HDF5 file only contains raw data, and it is it the 3D format.

psavery:
From my discussion with @cjh1:

To do: create an attribute on `/frames` in the HDF5 file that indicates the scan dimensions.
To do: try to make dimension attributes into tuples if possible, but separate attributes if not.

PR #133, opened 2020-02-19
Save the scan dimensions to /frames

psavery:
In the HDF5 format, save the scan dimensions to /frames as
a list of two attributes (width, height).

This also changes the convention we currently have to save
dimension attributes as a list of two attributes (width, height)
rather than as separate attributes, such as Nx and Ny.

This is not fully tested and may not fully work.

Fixes: #126

### DIFF
--- a/docs/source/StempyH5Format.rst
+++ b/docs/source/StempyH5Format.rst
@@ -17,7 +17,7 @@ Currently, the structure of the format is as follows:
    frames
 
 
-* 
+*
   ``electron_events/frames`` - Array of arrays, the first index is the scan position
   ( corresponding to ``scan_positions`` ) and the second array holds an array of
   of indices into the diffractogram where an electron strike was detected.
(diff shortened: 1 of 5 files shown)


psavery:
@cjh1 Unfortunately, I ran out of time to test this, but I decided to push it as it is. It is almost complete. But, I think that the writing of the `scan_dimensions` attribute in the sector reader might not work.

I think this PR is a step in the right direction, though.

ercius:
We should finish this up and update some of the HDF5 machinery to use the `scan_dimensions` attritbute on the 'frames` dataset.

1. `io.write_hdf5` adds `scan_dimensions` attribute to `frames`
2. `image.electron_count` uses the `scan_dimensions` attribute instead of the shape of `stem/images`
3. Check other places `stem/images` shape is used

psavery:
@cjh1 I believe this should work, although I did not test it.

cjh1 (approved):
LGTM

Issue #138, opened 2020-03-15
core dump for electron counting using latest version


ercius:
Im getting a core dump when doing electron counting using an HDF5 file for input. Im using the latest singularity image to run the counting.

Here is the code Im using:
``` python
from pathlib import Path
import sys
import time

import numpy as np
import h5py

import stempy.io as stio
import stempy.image as stim

scanNum = sys.argv[1]
scanName = 'data_scan{}_gain.h5'.format(scanNum)

threshold = int(sys.argv[2])

dPath = '/mnt/nvmedata1/temp'
fPath = Path(scanName)

iPath = dPath / fPath

dark0 = np.zeros((576,576))

print('Opening: {}'.format(iPath))
with h5py.File(iPath,'r') as f0:

    # Electron count the data
    sReader = stio.reader(f0)

    print('start counting')
    t0 = time.time()

    ee = stim.electron_count(sReader,dark0,number_of_samples=1200,
                                            verbose=True,threshold_num_blocks=20,
                                            xray_threshold_n_sigma=175,
                                            background_threshold_n_sigma=threshold)
```

The initial testing for the distribution mean and sigma finishes correctly and the verbose output looks correct. Then it goes directly to a core dump.

The electron counting works properly when using the sector reader. So, its probably something that has to do with the hdf5 reader specifically.

ercius:
@cjh1 Im uploading to Cori now. The file is:
.../ncemhub/4Dcamera/2020.03.13/data_scan16.h5

Its about 10GB.

cjh1:
Thanks @ercius copying now ...

psavery:
I was able to reproduce this, and I think I figured out the problem.

On the C++ side, we need to acquire the python GIL before we delete python objects. We weren't doing that in a couple of cases. I have a fix coming soon, and it fixes the crashing for me.

cjh1:
@ercius The image is update with the fix.

ercius:
awesome. Thanks so much!

PR #139, opened 2020-03-17
Ensure GIL is acquired before deleting py::array

psavery:
Although the DataHolder class was acquiring the GIL before
decrementing the reference count on the shared pointer when "reset()"
was called, there were a few cases where the GIL was not acquired.
This causes a crash when the last shared pointer is destroyed.

The cases in DataHolder where the GIL was not acquired were:

1. Destructor
2. Copy assignment
3. Move assignment

Both 1 and 2 were being performed in the electron counting code.

This hopefully fixes all of the cases where the shared pointer's
reference count is decremented without the GIL being acquired.

Fixes: #138

### DIFF
--- a/python/pyreader.h
+++ b/python/pyreader.h
@@ -14,6 +14,39 @@ namespace stempy {
 struct PYBIND11_EXPORT DataHolder
 {
   DataHolder() = default;
+  DataHolder(const DataHolder&) = default;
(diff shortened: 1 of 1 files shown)


psavery:
@cjh1 If you can think of any other cases where we might be destroying that shared pointer or decrementing its reference count without acquiring the GIL, let me know. But this should hopefully cover it.

Issue #140, opened 2020-03-17
feature request: read electron counts from h5 file


ercius:
Can we make the [ElectronCountedData](https://github.com/OpenChemistry/stempy/blob/1b3360fc114ddbc3a1ac9d70b5b461782bb523e9/python/stempy/image/__init__.py#L345) numedtuple available for use later? It would be great to provide the ability to be create a namedtuple which the user populates and stempy recognizes. Also, we can have a stempy.io function to load the electron counted data back into this same namedtuple form. Then the frame dimensions and scan dimensions will be placed correctly on import.

cjh1:
@ercius We can take a look at this. Providing a function to load it into the namedtuple should be pretty straight forward, we would need to think about how to allow a user to create and populate their own, but we will come up with something.

ercius:
@cjh1  thanks. I think that just loading the data back into the namedtuple would be sufficient. It should have the same signature as the originally created one so that it can be used in the stempy.image functions as well.

cjh1:
@ercius This is now merged into master and I have updated the container. Note that we changed `save_electron_counts` to take the `ElectronCountedData` rather than just the raw data, so you don't has to pass the parameter separately.

ercius:
That's exactly what I needed. Thanks! This makes analysis easier.

PR #141, opened 2020-03-24
Read electron counted data from HDF5 file

psavery:
Basically, you can now read the electron counted data from an
HDF5 file and then use it to create a stem image like so:
```python
counts = io.read_electron_counts('/path/to/file.h5')
stem_img = image.create_stem_images(counts, inner_radius, outer_radius)[0]
```

You can also now create your own python class that has the attributes
`data`, `scan_dimensions`, and `frame_dimensions`, and you can pass
that into `image.create_stem_images()` to be used as electron counted data.

Fixes: #140

### DIFF
--- a/python/stempy/io/__init__.py
+++ b/python/stempy/io/__init__.py
@@ -162,6 +162,31 @@ def save_electron_counts(path, events, scan_dimensions, frame_dimensions=None):
 
         frames[...] = events
 
+def read_electron_counts(path):
+    """Read electron counted data from an HDF5 file.
+
+    :param path: path to the HDF5 file.
+    :type path: str
+
+    :return: the coordinates of the electron hits for each frame.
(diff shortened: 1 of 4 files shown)


cjh1:
@psavery As part of this lets update save_electron_counts, so that it will also accept a ElectronCountData object ( so your not force to pass in the dimensions as well).

python/stempy/io/__init__.py
```diff
@@ -162,6 +162,31 @@ def save_electron_counts(path, events, scan_dimensions, frame_dimensions=None):
 
         frames[...] = events
 
+def read_electron_counts(path):
```
cjh1: I think this should be `load_electron_counts` ( as in save/load )

psavery: Done

psavery:
@cjh1 Done! This is ready for review again.

PR #142, opened 2020-03-26
sparse data processing functions and readme update

ercius:
1. I added three functions I commonly use for the sparse electron counted data. Each function works on only the sparse electron counted data and avoids expanding to the dense frame version. This saves on memory and processing.

- com_sparse(): Center of mass
- calculate_sum_sparse(): Create a dense frame from event data. Each set of events are added to the frame.
- radial_sum_sparse: Calculate a radial sum and return the data as a 3D ndarray. Makes interaction and creation of stem images quicker and simpler.

2. I also updated the readme file (#129) to include examples of actual processing. Can we add these images to the readme? Im not sure where we can store them so they can be referenced.

data_scan139_stemImages-2

data_scan139_dp


cjh1:
@ercius Great to see you first contribution! Will take look, looks like it might need rebasing on master, but we can take care of that.

cjh1:
@ercius I cherry-picked you changes onto current master and raise  #143, that fixes the issue introduced by the previous rebase.

ercius:
@cjh1 Glad to add my functions in so they can be used more broadly.  Any reformatting or suggested changes are welcome to match the existing code.

cjh1:
Closing in favor of #143

PR #146, opened 2020-05-07
Add support for contiguous data format

cjh1:
This add a need file format version (5) to support contiguous sector layout.

### DIFF
--- a/stempy/reader.cpp
+++ b/stempy/reader.cpp
@@ -478,33 +478,44 @@ Block SectorStreamReader::read()
   return Block();
 }
 
-void SectorStreamReader::readSectorData(Block& block, int sector)
+void SectorStreamReader::readSectorData(std::ifstream& stream, Block& block,
+                                        int sector)
 {
   if (version() == 4) {
-    readSectorDataVersion4(block, sector);
+    readSectorDataVersion4(stream, block, sector);
   } else {
-    readSectorDataVersion5(block, sector);
+    readSectorDataVersion5(stream, block, sector);
   }
 }
 
-void SectorStreamReader::readSectorDataVersion4(Block& block, int sector)
+void SectorStreamReader::readSectorData(Block& block, int sector)
+{
+  auto& sectorStream = *m_streamsIterator;
+  auto& stream = sectorStream.stream;
+
+  readSectorData(*stream.get(), block, sector);
+}
+
+void SectorStreamReader::readSectorDataVersion4(std::ifstream& stream,
+                                                Block& block, int sector)
 {
   auto frameX = sector * SECTOR_DIMENSIONS_VERSION_4.first;
   for (unsigned frameY = 0; frameY < FRAME_DIMENSIONS.second; frameY++) {
     auto offset = FRAME_DIMENSIONS.first * frameY + frameX;
-    read(block.data.get() + offset,
+    read(stream, block.data.get() + offset,
          SECTOR_DIMENSIONS_VERSION_4.first * sizeof(uint16_t));
   }
 }
 
-void SectorStreamReader::readSectorDataVersion5(Block& block, int sector)
+void SectorStreamReader::readSectorDataVersion5(std::ifstream& stream,
+                                                Block& block, int sector)
 {
   auto frameY = sector * SECTOR_DIMENSIONS_VERSION_5.second;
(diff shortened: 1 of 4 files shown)


ercius:
What's the status of this pull request? We can test it tomorrow if there is a singularity image to pull.

cjh1:
@ercius This should be ready to go, it would be great if you can test it. I will push a new image.

cjh1:
`openchemistry/stempy:latest` is now built from this branch. Let me know if you run into any problems.

stempy/reader.cpp
```diff
@@ -278,9 +278,22 @@ int extractSector(const std::string& fileName)
...
 }
 
-SectorStreamReader::SectorStreamReader(const vector<string>& files)
-  : m_files(files)
+SectorStreamReader::SectorStreamReader(const vector<string>& files,
+                                       uint8_t version)
+  : m_files(files), m_version(version)
 {
+  // Validate version
+  switch (this->m_version) {
```
cryos, line 286: When using m_ for member variables it would be preferable to avoid using this-> as well

stempy/reader.cpp
```diff
@@ -278,9 +278,22 @@ int extractSector(const std::string& fileName)
...
+    case 4:
+    case 5:
+      break;
+    default:
+      std::ostringstream ss;
+      ss << "Unsupported version: ";
+      ss << this->m_version;
```
cryos, line 293: As above

ercius:
Thanks! I'll test it today.

ercius:
I just tried this on the test data 2020.04.29/data_scan5*.data but on the processing computer. Its not reading the files correctly. When trying to do electron counting I get these results for the fitting:
```
****Statistics for calculating electron thresholds****
number of samples:1200
min sample:-32768
max sample:32256
mean:115.434
variance:354272
std dev:595.208
number of bins:65024
x-ray threshold n sigma:175
background threshold n sigma:4
optimized mean:9.40402
optimized std dev:7.70378
background threshold:40.2191
xray threshold:104277
Segmentation fault (core dumped)
```
The min and max are outside uint16 values. This is with the backend='thread' option.

Without that option I get this error:
```python
Traceback (most recent call last):
  File "/home/daquser/temp_pae/electron_count.py", line 46, in <module>
    background_threshold_n_sigma=threshold)
  File "/usr/local/lib/python3.7/site-packages/stempy/image/__init__.py", line 324, in electron_count
    blocks.append(next(reader))
  File "/usr/local/lib/python3.7/site-packages/stempy/io/__init__.py", line 19, in __next__
    b = self.read()
  File "/usr/local/lib/python3.7/site-packages/stempy/io/__init__.py", line 31, in read
    b = super(ReaderMixin, self).read()
ValueError: Unexpected EOF while processing stream.
```
Can you test that your container works when you pull stempy:latest using:
`singularity pull docker://openchemistry/stempy:latest`

Im calling the electron counting with these lines:
```python
sReader = stio.reader(iFiles,stio.FileVersion.VERSION5, backend='thread')
ee = stim.electron_count(sReader,dark0,number_of_samples=1200,
                                            verbose=True,threshold_num_blocks=20,
                                            xray_threshold_n_sigma=175,
                                            background_threshold_n_sigma=threshold)
```

cjh1:
@ercius Can you test removing the `backend='thread'`?

cjh1:
I ran the test myself, seems to work in serial mode, so I will do some digging ....

ercius:
I just tested without the backend='thread' also. That works. I made a mistake in my earlier test where I removed the fileversions and the backend keyword. 

So, the only issue is with the threaded reader and the contiguous format.

ercius:
This is working now. Im using the stempy-threaded singularity image with the threaded option. So, this can be added to master.

cjh1:
@ercius Excellent, thanks for testing.

cjh1:
@ercius This is now merge, I have updated in the images ( including the images on cori ).

ercius:
Awesome. Thanks for getting that all set up. I'll test Cori when I get a
chance.

ercius:
Tested on Cori and it's working well. Threading on Haswell nodes are only 25% slower than our local analysis PC.

cjh1:
That is great news. Was that from a burst buffer? Interactive or batch?

cjh1:
@ercius I just saw the email, thanks.

PR #159, opened 2020-07-24
Improve speed, update examples, add com_dense function

ercius:
1) Improve the speed of the com_sparse function
2) Update the examples with code we use on the local event building PC
3) Add a com_dense function to help with analysis of raw and counted but dense frames

### DIFF
--- a/examples/create_images.py
+++ b/examples/create_images.py
@@ -3,42 +3,18 @@ from pathlib import Path
 
 import stempy.io as stio
 import stempy.image as stim
-import numpy as np
-import h5py
-
-import imageio
 
 scanNum = sys.argv[1]
 
-scanName = 'data_scan{}'.format(scanNum)
-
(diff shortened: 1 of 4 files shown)


cjh1:
@ercius Do you mine if I parameterize the paths, as in allow them to be passed on the command line, so they aren't hardcoded?

ercius:
Please do!

cjh1:
@ercius I have added a few more command line options using click ( `pip install click`). The scripts should work just as they did fore you ( using the hardcoded path ) , but you can override the paths and they will also print nice usage information for you. Let me know if you have any problems.

ercius:
@cjh1 Is click installed in the docker images? I use the code through those so it should work as long as click is already in there. 

Thanks for cleaning that up! Its useful to be able to set the paths and get all the extra information. Good to merge once its reviewed.

cjh1:
Yes, `click` is installed in our containers. I will go ahead and merge.

Issue #154, opened 2020-07-01
Operating on electron events slow when loaded from disk


ercius:
There is a big difference in operating on data returned by stempy.electron_count() and stio.load_electron_counts(). I tested this using stim.create_stem_images(). The attached PDF shows a notebook with the output. The results show a difference in 0.07 seconds vs 4 seconds.

Is there a way to make the loaded data in the same form as the data returned by the electron_counting code?

The data and notebook are available on ncemhub on Cori:
../ncemhub/4Dcamera/stempy_count_load_test.ipynb

[stempy_count_load_test.pdf](https://github.com/OpenChemistry/stempy/files/4859088/stempy_count_load_test.pdf)

cjh1:
The reason for this different I think is in the first case the memory is all allocated and use on the C++ side. In the second case we use h5py to load the data so it allocated by numpy, when we operate on it we have to go through the Python API. There are a couple of things we could try, such a loading directly with hdf5 ( we now it as a dep ).

ercius:
That was what I thought. If we can get similar speed for the loaded data that would be very useful for analyzing these data sets. Can you take a look at that?

cjh1:
Absolutely, I take a look.

ercius:
@cjh1 After the gain is merged it would be great to work on this. Thanks!

cjh1:
Absolutely, this is next on the list.

cjh1:
Fixed by #160

PR #160, opened 2020-07-27
Prevent copying numpy array electron counting data

cjh1:
Use py::array_t instead.


cjh1:
Fixed #154

cjh1:
@ercius I updated the jupyter lab image on cori, if you want to try these changes out.

ercius:
@cjh1 This is working. Its just as fast as when the data comes from the counting function. Thanks!

cjh1:
@psavery Could you do a quick review if you get a change?

cjh1:
@psavery Could you review this one as well?

Issue #132, opened 2020-02-19
Maybe add documentation for reader versions


psavery:
Currently, they are just `FileVersion.VERSION1`, `FileVersion.VERSION2`, etc. It might be good to add some documentation.

ercius:
@psavery @cjh1  This is a good idea. Where would documentation like this live? I can add descriptions of the headers. I already have them written down.

cjh1:
I think we could document it in a new rst file [here](https://github.com/OpenChemistry/stempy/tree/master/docs/source). If you have the description to had that would be great, we can also get it from the code. We can then add some description of the various versions, which might involve a little archaeology.

PR #166, opened 2020-08-10
Description of raw 4D Camera data files

ercius:
Add a description of VERSION3, VERSION4, and VERSION5 data types for raw 4D Camera files.

addresses #132


docs/source/4dCamera_header_versions.rst
```diff
@@ -0,0 +1,141 @@
+
+NCEM 4D Camera Raw Data Versions
+================================
+
+This is a record of the layout of raw 4D Camera data. It inlucdes
```
cjh1, line 5: inlucdes => include

PR #174, opened 2020-09-18
Use column rather than index

cjh1:


cjh1:
@psavery I think I have found it!

psavery (approved):
LGTM

ercius:
@cjh1 Nice! Can I test it by pulling the latest container?

cjh1:
@ercius Yes, I have pushed a new image.

PR #175, opened 2020-09-18
Fix maximalPointsParallel templates

cjh1:
It appears some issues where introduce in the last branch. Ensure that we pass the frame type down to VTK-m


psavery (approved):
LGTM

Issue #157, opened 2020-07-16
Validate numpy types, passing the wrong type currently causes a crash


cjh1:

cjh1:
@psavery I think this can be closed off, given the work you did in #178

psavery:
@cjh1 Sounds good. Yeah, it is ensuring the `gain` and `darkreference` variables are cast to the right type. We can use that utility function more in the future if more things need to be cast.

cjh1:
Fixed in #178

PR #178, opened 2020-10-02
Use 32-bit floats for dark reference type

psavery:
Instead of a 64-bit double, use 32-bit floats for the dark reference
type, since 32-bit is usually big enough.

Fixes: #156

### DIFF
--- a/python/image.cpp
+++ b/python/image.cpp
@@ -20,7 +20,7 @@ namespace stempy {
 
 template <typename BlockType>
 CalculateThresholdsResults<uint16_t> calculateThresholds(
-  std::vector<BlockType>& blocks, py::array_t<double> darkReference,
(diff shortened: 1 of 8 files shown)


psavery:
This crashes if the user passes numpy float64 instead of float32 as the dark type. This is the same kind of issue as #157.

psavery:
Here are before and after pictures for an example I have, where we made the dark background 1.5. They look similar.

image

image

psavery:
@cjh1 Ready for review

cjh1 (approved):
LGTM, thank you

cjh1:
I have updated the image.

ercius:
Thanks! It looks good. I'll see if I can come up with a good quantitative
test. But I think we can call this done.

Issue #176, opened 2020-10-01
Multiple frames per probe position


ercius:
@cjh1 Do know how the counting code will deal with having two frames acquired at the same probe position? This will allow us to improve signal by adding multiple frames together. Im not sure if this was incorporated into the raw data reader though.

cjh1:
So currently the assumption is one frame per probe position. The current behavior would be that the electron counts would be from the last frame. What are your thoughts on how combine the multiple frames?

ercius:
I think I would just append the second counted frame to the same STEM scan position. If the two frames have electrons at the same position then the list of positions will just have the same number twice. I believe that the analysis would treat that the same and nothing would have to change downstream (like create_stem_images()). Does that sound good?

cjh1:
Yes, that makes sense, it should be a pretty simple change.

ercius:
ok...I think that would be worth trying. I took a data set with 2 frames per position. Its just noise unfortunately, but Ill try to get one with a real sample soon.

ercius:
@cjh1 The expected behavior unfortunately does not happen (ignoring the first frame(s)). I get a core dump when I try to count a file with multiple frames per position.

cjh1:
:disappointed: If you point me at the dataset I will take a look.

ercius:
No worries! We never tried it before. Actually, there are two different behaviors.

- If I use 4 frames per position then I get a core dump when I try to count. Not sure why. Scan #225
- If I use 2 frames per position then it counts. I guess it throws the first frame away. So, that is as expected! Scan #227

The data is on /mnt/nvmedata*/ right now. They are both 129x128 scans. There are also two similar scans with 1 frame per position (224 and 228)

cjh1:
OK, I will grab scan225, will let you know when I have it downloaded.

ercius:
This issue can probably be figured out when we implement the full capability to count multiple frames per position. Counting that data set can be left to later. Its not critical. It is good to know that >2 frames crashes things though.

cjh1:
@ercius So I am able to recreate the segfaults, lots of them :smile: So this isn't quite as simple as I first thought. One question I have is how are the multiple frames at a single position identified?  We use the scan position as the key when reconstruction the partial frames. So the question is how do we know which parts fit with with frame, is there some differentiation in the header? The scan number?

ercius:
@cjh1 Great question. I had not thought of that.

[...]

PR #179, opened 2020-10-07
Add support for multi frame scans

cjh1:


cjh1:
@ercius This should fix the issue we are seeing. It certainly does with scan225. The events from any addition frames are now append rather than replacing the previous frame. Note there made be a small performance hit, as we now need to protect that append operation from access by multiple threads, but it shouldn't be too bad. I have push as test image that you can try out: 

```
singularity pull docker://openchemistry/stempy:multi-frame
```

cjh1:
Fixes #176

ercius:
This works! Awesome work. That is super useful. Good to merge.

psavery (approved):
LGTM

ercius:
Great! Can we update the images? I'll pull it to mothership6

cjh1:
Done, new image is uploaded.

Issue #172, opened 2020-08-22
Out of memory on Cori nodes


ercius:
Getting out of memory issue on Cori when counting 163 GB (512x512) sized scan.  Using 20 threads the nnode runs out of memory. Using 15 threads is successful.

Same problem with 20 threads when using `gain=None` and `gain=np.ones((576,576),dtype=np.float32)`

Would be good to check the dtype used in memory for different options:
- [ ] No gain and no dark
- [ ] Dark, no gain
- [ ] Gain and dark

cjh1:
You mentioned that this is with `VERSION4`, is this correct?  

So in the case of the first two that data will be treated as `uint16`, only the introduction of the gain will push the size to `float32`. In the case of subtracting the dark reference, we statically cast back to `unit16` ( truncate ) which may or may not be the right think to do? However, this is what we have been doing from the start. We have an issue that we will work on next to accept `float32` for the dark reference as well.

So in terms of the out of memory issue. If this is running on a knl node we have 96gb to play with. The main memory usage for the threaded reader is to cache the partial frames as they reconstructed ( they are removed from the cache and counted once they are complete ). So there are a couple of things that could cause problems in terms of the dataset. If there is large loss ( lots of partial frames ) or if the frame quadrants are for some reason jumbled across the multiple files ( as in we have to maintain a lot of partial frames for the entire read ). How many files is this data set spread across? In general more threads is also going to increase the memory usage as you are seeing.

ercius:
> You mentioned that this is with VERSION4, is this correct?

Yes. And if that makes a difference then we should not bother fixing this. I'm only interested in VERSION5 going forward. I was using an older experiment as an example of a folder full of data that was not fully analyzed and Im testing submitting many counting jobs at once.

>So in the case of the first two that data will be treated as uint16, only the introduction of the gain will push the size to float32. In the case of subtracting the dark reference, we statically cast back to unit16 ( truncate ) which may or may not be the right think to do? However, this is what we have been doing from the start. 

Everything you describe here is what I would expect the code to do. I was getting the problem on Cori KNL nodes and not on mothership6. Probably one has more memory than the other.

> We have an issue that we will work on next to accept float32 for the dark reference as well.

I dont think there is any need for this. Dark can be uint16 and gain should be float32.

I think that the only thing I wanted to do with the dark was to have the input allow uint16. Right now I think it requires it to be float64 (default of np.zeros()). Sorry if I was not clear on that particular point.

[...]

PR #183, opened 2020-11-03
Release stream before processing frame

cjh1:
By holding on to the stream while processing a frame we were cause starvation of that stream. This caused the streams to be read unevenly cause the cache to grown as frames could not be reconstructed.

Fixes: #172

### DIFF
--- a/stempy/reader.cpp
+++ b/stempy/reader.cpp
@@ -770,25 +770,25 @@ void SectorStreamThreadedReader::initNumberOfThreads()
   }
 }
 
-bool SectorStreamThreadedReader::nextSectorStreamPair(
-  SectorStreamPair& sectorStreamPair)
+bool SectorStreamThreadedReader::nextStream(StreamQueueEntry& streamQueueEntry)
 {
   {
     std::unique_lock<std::mutex> queueLock(m_queueMutex);
     if (m_streamQueue.empty()) {
       return false;
     }
-    sectorStreamPair = m_streamQueue.front();
+    streamQueueEntry = m_streamQueue.top();
     m_streamQueue.pop();
   }
 
-  auto& stream = sectorStreamPair.first;
+  auto stream = streamQueueEntry.stream;
 
   auto c = stream->peek();
   // If we have reached the end close the stream and remove if from
   // the list.
   if (c == EOF) {
     stream->close();
+    std::unique_lock<std::mutex> streamsLock(m_streamsMutex);
     auto iter = m_streams.begin();
     while (iter != m_streams.end()) {
       if ((*iter).stream.get() == stream) {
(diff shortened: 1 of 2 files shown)


cjh1:
@ercius This should resolve  #172. I have built and pushed a image to cori to test with: 

```
openchemistry/stempy:debug
```

You shouldn't need to specify the number of threads, it all the threads available, which it will do be default.

cjh1:
@ercius I am able to count `ncemhub/4Dcamera/2020.10.05/data_scan0000000231` in about 4 minutes 20 seconds.

ercius:
Exciting! I will try this right away.

ercius:
I got a data set of 512x512 to count in about 513 seconds (8.5 minutes). In

I then tried the one you were using to test which finished in 289 seconds.
So, that other one just had more electrons.

Success!

It's about 4x slower than our local buffer, but we can count many say once.

Did you try a 1kx1k?

ercius:
I got a seg fault on a different data set. I dont have time to look into it right now. Ill run it again tomorrow and make sure I didnt make weird mistake (like providing non-data files to the reader).

ercius:
It was bothering me and I looked at it anyway. I re-ran the code and it **worked fine.** Not sure why. False alarm most likely.

cjh1:
@ercius Didn't try a 1k x 1k will be interested in how it performs, as I think with full number of thread on a knl node we have extra capacity that would get used in that case. If you see the seg fault again let me known, it could well be a timing issue that you only see sometimes, given the number of threads we are now using.

psavery (approved):
LGTM

ercius:
@cjh1 Unfortunately, the 1kx1k scan I tried to count ran out of memory before finishing. Im not sure how far it got before quitting.

cjh1:
@ercius Can you point me to the dataset and I will take a look.

ercius:
2020.06.24/data_scan000000051_*.data

I ran it from my scratch. It ran for a while (longer than it used to run before your fix) but still overwhelmed the memory. I hope I was using the correct image (stempy:debug)

cjh1:
@ercius I have the data stage, but it appears cori is "Degraded" at the moment so I can run any jobs ...

ercius:
I hope we can get back to it soon. Getting this to work would be excellent.

cjh1:
Just a quite update on this. I have made some good progress. I can now count the scan without running out of memory. However, it appear to running out of memory when I copy the counts back to python. I need to do a little more investigation and may be some rework on the python side.

cjh1:
@ercius We have made some fixes and improvements, we are now about to count that dataset without running out of memory on cori. I have updated the test image (`openchemistry/stempy:debug`).

psavery (approved):
LGTM

ercius:
@cjh1 

> we are now about to count that dataset

How did the counting of the large dataset go? I can try it as well on a couple other ones when I get a chance.

cjh1:
@ercius Sorry we are able the count that data set now, it took about 15 minutes. It would be good to try with a couple of other datasets as well.

PR #187, opened 2020-11-20
Move electron data in no-copy operation to numpy

psavery:
This steals the data from the C++ vectors and moves it to numpy arrays.
The numpy arrays will automatically clean up the data when they are
finished.

This prevents a copy being made of the electron counted data.

### DIFF
--- a/python/image.cpp
+++ b/python/image.cpp
@@ -18,6 +18,36 @@ using namespace stempy;
 // we may want to rethink in this is in the future.
 namespace stempy {
 
+template <typename T>
+py::array_t<T> vectorToPyArray(std::vector<T>&& v)
+{
+  // Steals the vector pointer and returns a numpy array with the data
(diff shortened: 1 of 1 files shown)


python/image.cpp
```diff
@@ -18,6 +18,35 @@ using namespace stempy;
...
+}
+
+struct ElectronCountedDataPyArray
+{
+
+  ElectronCountedDataPyArray(ElectronCountedData&& other)
+  {
+    data.reserve(other.data.size());
+    for (auto& vec : other.data)
+      data.push_back(vectorToPyArray(vec));
```
cjh1, line 38: Can we add braces please

psavery: I fixed this and force-pushed.

cjh1 (approved):
LGTM, works well even with large dataset.

psavery:
Great! :slightly_smiling_face:

PR #191, opened 2021-01-12
Add first pass at SparseArray class with examples

psavery:
The SparseArray class can take electron counted data and use it to
perform operations such as min/max/mean/sum, indexing, and slicing.

Slicing and indexing operations are allowed, such as `array[3:60, 5, ::2, :3]`,
where the only the frames that get expanded are the ones that are needed.

Arithmetic operations can be performed, such as `np.sum(array, axis=(0, 1))`.
Note that only axis=(0, 1) is currently specialized for efficiency. If you
use an axis that isn't specialized, it is like calling
`np.sum(array[:], axis)`, where the array gets expanded to a full numpy array
and then the operation performed on it. We can specialize more axes options
as needed.

The class also accepts an argument that indicates whether or not full expansion
(such as `array[:]`) is allowed. If `False`, then an exception will be raised
if a full expansion is attempted. This is helpful to ensure that functions
that are called do not attempt a full expansion.

The class also accepts a `dtype`, which is the type that will be used when
an expansion happens. For things like a sum, it is important to pick a dtype
large enough so that it won't overflow.

The examples are taken from the py4dstem quick start and modified slightly to
use our SparseArray class. Some things to note:

1. We are using sum() instead of max() because max() for our example is mostly
   all ones with a few zeros (not very visually appealing).
2. The binning and cropping steps are removed. They work, but they do so in
   py4DSTEM by slicing and re-assigning (such as
   `datacube.data = datacube.data[:, :, 3:20, 4:15]`). Since slicing the sparse
   array returns a numpy array, this means the sparse array will get replaced
   by a numpy array, which we might not want for this example.
3. The saving to HDF5 at the end doesn't work because py4DSTEM doesn't
   recognize our object type.
4. We have to swap scan dimensions for the full battery example for the image
   to come out correctly. We might need to make sure we are using the dimensions
   correctly.

The two jupyter notebooks with their output are attached in this [examples.zip](https://github.com/OpenChemistry/stempy/files/5805251/examples.zip) file.

Let us know what you think, @ercius


ercius:
This looks great.  Ill try it out soon.

> The class also accepts an argument that indicates whether or not full expansion
(such as array[:]) is allowed. If False, then an exception will be raised
if a full expansion is attempted. This is helpful to ensure that functions
that are called do not attempt a full expansion.

I was really worried about the possibility to nuke the users RAM, but this is a great way to avoid this. Thanks for putting this in there. `False` should be the default value.

psavery:
@ercius Okay, I just force-pushed. It is now `False` by default.

ercius:
@psavery Is there a docker image I can download to test this? The example notebooks look like they are just what we want to accomplish. The dataset has insufficient signal for the Bragg disc detection. I will have to get a different dataset to test for that or try out some different py4dstem functionality.

cjh1:
@ercius Its really only a single python file ( `spare_array.py` ), it currently does need any of the other stempy machinery. So you can either get the file and notebook from the PR or we can send it to you?

ercius:
I can grab it. Thanks!

ercius:
@psavery This works really well. Nice work! Should we discuss the next steps to improving this and what to add?

cjh1:
@ercius Glad this works for you. Yes, we should discuss what other things to add and where the code should end up living ...

ercius:
Those are good questions.

More features to add are:
1. binning the diffraction data to change the frame dimensions in place. 
2.  cropping the data in both diffraction and real space
I have python code for both of these which I can share.

As for where it should live: I would like it to be easily accessible and installable. We should talk about how to make stempy more accessible or maybe to start a different repo or add it to an existing repo (ncempy?). Im not sure of the right way to do this and would like to discuss that.

cjh1:
@ercius Just so we are on the same page. For 1. and 2. these are operations supported in py4DSTEM? As in we would like to support them with our sparse array? Sharing the code would be great.

I agree that making stempy more accessible would be great. I know we have talked about pushing packages to pypi in the past but put it off because of the potential time sink. @psavery has setup several projects recently to build for pypi and I think we have streamlined the process. Given this it may be worth revisiting this for stempy. I think with a few hours investment we would know pretty soon if its going to be possible in a tractable amount of time. If this didn't work out the sparse array code could also end up in py4DSTEM itself.

ercius:
For 1 and 2 you can use functions in py4dstem, but I think it would require working on the dense data. I was planning for 1 to add a function to bin the data in place. You can change the way the sparse --> dense operation happens in order to reduce the frame_dimensions. It would be similar to how numpy arrays can be reshaped by only changing the shape and not actually changing the underlying data. For 2 I think this would also be built in but probably would need to return a new array. Im not sure how I would want that to work.

Lets see if we can get it distributable in pypi. That is certainly worth some effort at this point.

psavery:
Hi Peter @ercius,

I've added cropping (via slicing) and binning for frames. I've attached
[an example notebook with output](https://github.com/OpenChemistry/stempy/files/5991875/cropping_binning_with_output.ipynb.zip) and more details are below. Let us
know what you think.


Changes:
========

1. The SparseArray class now has an attribute called "sparse_slicing",
   which defaults to True. If True, then slicing a SparseArray will
   return a new SparseArray with the underlying data modified (except
   when the slicing results in a single frame, in which case it returns
   a dense array for the frame).

   This allows for cropping and striding the data via slicing. If
   "sparse_slicing" is False, then a dense array will be returned upon
   slicing.

   Since py4dstem also crops data via slicing, cropping a SparseArray
   with their functions should yield the same result.

2. Previously, the math functions and expansion functions assumed that
   the presense of a sparse value in a frame indicated that the pixel
   value is 1. However, this is not the case, because the frame can
   have duplicated sparse values, which indicates a higher pixel value.
   The underlying math and expansion functions have now been modified
   to reflect this, and they now count the number of sparse values in
   all cases. This is important also for binning, where duplicated
   values are needed to indicate greater intensity in a pixel.

   This results in a little bit of a slowdown for the functions, but
   accuracy is more important. Current timings are pasted below.

3. A bin_frames() function was added to bin the frames. This either
   returns a new SparseArray (default), or if "in_place" == True,
   it modifies the SparseArray in place and returns "self".

   We currently can't use the py4dstem functions to bin the data,
   because they perform a reshape and sum. We can only bin the frames
   via this "bin_frames()" function.


Questions:
==========

1. There is a dtype in the SparseArray class which indicates what
   dtype should be used for expansion and math functions. This
   currently defaults to np.uint8, which causes integer overflows
   for most of our examples, for things like summing. Should we
   increase the default to something else, like np.uint32?

2. Adding a "bin_scans()" function might not be too hard. Should we
   add one of those too, to bin the scans?


Current math timings:
=====================

Small dataset:
data.frame_dimensions=[576, 576]
data.scan_dimensions=[64, 64]

min:  3.90 s
max:  0.92 s
sum:  0.86 s
mean: 0.84 s


Large dataset:
data.frame_dimensions=[576, 576]
data.scan_dimensions=[257, 1024]

min: 233.35 s
max:  54.15 s
sum:  50.57 s
mean: 49.81 s

ercius:
@psavery  Here are some answers:
> There is a dtype in the SparseArray class which indicates what
> dtype should be used for expansion and math functions. This
> currently defaults to np.uint8, which causes integer overflows
> for most of our examples, for things like summing. Should we
> increase the default to something else, like np.uint32?

Yes. I would definitely increase to uint32. uint8 is too small. Thanks!

> Adding a "bin_scans()" function might not be too hard. Should we
> add one of those too, to bin the scans?
Im not sure what bin_scans() would do. I thought the binning was already done in the array? A little more information is needed.

psavery:
@ercius Sounds good about increasing the default dtype to `uint32`.

Py4dstem has separate functions for binning the frames ([`bin_data_diffraction`](https://github.com/py4dstem/py4DSTEM/blob/3551d98a6930a0ca3e678c57440da76b28b160b5/py4DSTEM/process/preprocess/preprocess.py#L77)) and binning the scans ([`bin_data_real`](https://github.com/py4dstem/py4DSTEM/blob/3551d98a6930a0ca3e678c57440da76b28b160b5/py4DSTEM/process/preprocess/preprocess.py#L117)).

For a 4D shape, binning the scans would perform the binning on the first 2 dimensions, whereas binning the frames would perform the binning the last 2 dimensions, if I understand it correctly.

For our sparse array, I think we could just concatenate some of the internal sparse frames to achieve this.

We only really need to add this, though, if you think it'd be useful.

psavery:
The default dtype is now uint32.

ercius:
@psavery 

> For our sparse array, I think we could just concatenate some of the internal sparse frames to achieve this.

Yes. This would be very useful, and is the method I would also suggest using. Good idea!

psavery:
@ercius Alright! `bin_scans()` has been added!

psavery:
The sparse array was moved to `stempy.io.sparse_array`.

It can now be imported via:

```python
from stempy.io.sparse_array import SparseArray
```

python/stempy/io/sparse_array.py
```diff
@@ -0,0 +1,443 @@
...
+def default_full_expansion(func):
+    name = func.__name__
+
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        ret = func(self, *args, **kwargs)
+        if ret is not None:
+            return ret
+
+        print(f'Warning: performing full expansion for {name}')
```
cjh1, line 29: Lets write warnings to stderr.

cjh1 (approved):
@psavery This is looking great. Just one inline suggest. Lets add a little documentation, it can be pretty rough and ready, may just cut and paste some of the conversion from the PR, so we at least have a starting point.

cjh1:
@ercius I think this is pretty much ready to merge, if you are OK with that. We can of course do follow up PRs.

python/stempy/io/sparse_array.py
psavery: This is fixed.

psavery:
I quickly went through it and wrote some doc strings. It didn't take very long. I also tried to get it to be added automatically to read the docs. Hopefully that will work...

cjh1 (approved):
@psavery Docs looks good, thanks. @ercius If its OK, we will go ahead and merge this. Any issues can be fixed in the follow up PRs. Once its in I can cut another patch release so you can install from PyPI. @psavery Is there away we could push commits to master to PyPI on a dev channel, so we could install master for testing?

cjh1:
@psavery I guess I could also just push a pre-release tag ...

ercius:
This is all great. Thanks!

cjh1:
@ercius As this has been merge I cut a pre-release, you install using pip like this:

```
pip install --pre stempy
```

Note: The docs for it are also available:

https://stempy.readthedocs.io/en/latest/stempy.io.html

ercius:
@psavery I was trying to install this preview version to test it for some users before I send them data. But it looks like `sparse_array.py` is not included in the release.

I tried `pip install --pre stempy` and `pip install stempy==1.1.0a1` but the `SparseArray` class is not available.

If you go to the [download page on pypi](https://pypi.org/project/stempy/1.1.0a1/#modal-close) and download the wheel you can open it like a zip file. The linux wheel does not contain the `sparse_array.py` at `stempy-1.1.0a1-cp36-cp36m-win_amd64.whl\stempy\io\`

We probably should just release `1.1.0`. I have used the `SparseArray` class and it works well so far.

cjh1:
@ercius Yes, we can go ahead and cut a release.

cjh1:
@psavery Any idea what is going on here?

psavery:
I believe that got fixed in #205, which was merged after the `1.1.0a1` tag. I believe it should be in the `1.1.0` release when it is uploaded.

psavery:
@ercius @cjh1 Indeed, stempy `1.1.0` was just uploaded to PyPI, and it contains the `SparseArray` class in `stempy.io.sparse_array`! I am able to import it.

PR #192, opened 2021-02-02
Add multi-pass reader

cjh1:
This readers the headers first to build up a frame map and then uses a second pass to read the data. This reduces memory usage and will provide the opportunity to parallelize across nodes.

### DIFF
--- a/stempy/reader.h
+++ b/stempy/reader.h
@@ -201,7 +201,9 @@ public:
   {
     std::unique_ptr<std::ifstream> stream;
     int sector = -1;
-    SectorStream(std::ifstream* str, int sec) : stream(str), sector(sec){};
+    // Mutex to guard access to the ifstream
+    std::unique_ptr<std::mutex> mutex;
+    SectorStream(std::ifstream* str, int sec) : stream(str), sector(sec), mutex(std::make_unique<std::mutex>()){};
   };
 
 protected:
@@ -304,7 +306,7 @@ public:
   template <typename Functor>
   std::future<void> readAll(Functor& f);
 
-private:
+protected:
   // The number of threads to use
   int m_threads = -1;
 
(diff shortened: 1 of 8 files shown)


stempy/reader.h
```diff
@@ -444,6 +448,165 @@ std::future<void> SectorStreamThreadedReader::readAll(Functor& func)
...
+// to build up "frame maps" detailing the streams and offsets for all the
+// sectors in a frame. The second pass is used reconstruct the frames using this
+// map. This avoids the need to cache partial frames so used less memory.
+// Depending on the seek performance of the disks it also performs well.
+class SectorStreamMultiPassThreadedReader : public SectorStreamThreadedReader
+{
+public:
+  SectorStreamMultiPassThreadedReader(const std::string& path);
+  SectorStreamMultiPassThreadedReader(const std::vector<std::string>& files);
+  SectorStreamMultiPassThreadedReader(const std::string& path, int threads = 0);
```
psavery, line 480: I think the default argument `threads = 0` would never get used, since the single argument `SectorStreamMultiPassThreadedReader(const std::string& path);` would get preference in overload resolution.

Would it make sense to just have a single `SectorStreamMultiPassThreadedReader(const std::string& path, int threads = -1);` constructor? Also, similar kind of thing for the vector version.

stempy/reader.h
```diff
@@ -444,6 +448,165 @@ std::future<void> SectorStreamThreadedReader::readAll(Functor& func)
...
+    m_futures.emplace_back(m_pool->enqueue(
+      [this, &func, header]() { processFrames(func, header); }));
+  }
+
+  // Return a future that is resolved once the processing is complete
+  auto complete = std::async(std::launch::deferred, [this]() {
+    for (auto& future : this->m_futures) {
+      future.get();
+    }
+  });
```
psavery, line 606: Very nice!

stempy/reader.h
cjh1: Yes, that makes sense, I have pushed a fix.

psavery (approved):
LGTM

PR #193, opened 2021-02-19
Add support for electron counting across nodes using MPI

cjh1:
This PR builds on the multi pass reader to perform electron counting across nodes. The driving python script is pretty much the same, you just need to remember that a copy of the script will be run on each node. The electron event are only gather to rank zero. So when writing out the data you need todo something like this:

```
from mpi4py import MPI

comm = MPI.COMM_WORLD
rank = comm.Get_rank()

# Usual reader logic goes here
...

if rank == 0:
    print('total time = {}'.format(t1-t0))
    stio.save_electron_counts('/global/cscratch1/sd/cjh1/electrons.h5', ee)
```

That way we write out the results on rank 0.

Here is an example submission script ( the XXX would need to be replaced ):

```
#!/bin/bash
#SBATCH --qos=debug
#SBATCH --time=00:30:00
#SBATCH --nodes=16
#SBATCH --constraint=haswell
#SBATCH --account=XXX
#DW jobdw capacity=1000GB access_mode=striped type=scratch
#DW stage_in source=/global/cscratch1/sd/XXX/scan000000051 destination=$DW_JOB_STRIPED/scan51 type=directory
echo "Number of node: $SLURM_JOB_NUM_NODES"
srun -N $SLURM_JOB_NUM_NODES shifter --image=openchemistry/stempy:multi-mpi -- python3 ~/4dstem/electron_count_multi.py $DW_JOB_STRIPED/scan51
```

The results are pretty good ( this is using BurstBuffers ):

scan126 ( 512x513 ) can be counted in 17 second over 16 nodes. 
scan51 ( 1k X 1k ) can be counted in 40 seconds over 16 nodes

Even with just a few nodes such as 4 we get good speed up compared with a single node, depending on scan size. We should do some more detailed scaling runs.


stempy/reader.h
```diff
@@ -2,10 +2,14 @@
 #define stempyreader_h
 
 #include "config.h"
+<<<<<<< HEAD
+=======
```
psavery, line 6: @cjh1 Right here

CMakeLists.txt
```diff
@@ -51,6 +70,15 @@ set(_src
...
+if (stempy_ENABLE_MPI)
+  list(APPEND _src
+    stempy/electron_mpi.cpp
+  )
+  list(APPEND _src
+    stempy/reader_mpi.cpp
+  )
+endif()
+
 set(_libs Threads::Threads)
```
psavery, line 82: One more thing! I think we should remove this extra `set(_libs Threads::Threads)` line. It's over-writing the `_libs` that was set earlier to include MPI!

That may also be why linking to the MPI target isn't including the header files...

psavery (approved):
After https://github.com/OpenChemistry/stempy/pull/193#discussion_r586081162 is fixed, LGTM

CMakeLists.txt
cjh1: Yes, I think that fixed it!

psavery (approved):
LGTM

PR #194, opened 2021-02-19
Build and test wheels

psavery:
This is a start for wheel building on github actions. Linux wheels for python 3.6 - 3.9 are built and tested with a simple test.

Some of the things this introduces:

1. Use of scikit-build in the setup.py file, which gets setup.py to run cmake
2. pyproject.toml to specify dependencies for running setup.py
3. Only find Development.Module for python cmake (requires cmake 3.18 - we can backport the module if this is too new - I have another repo where I have done this)
4. Add FindEigen3.cmake (copied from openchemistry/avogadrolibs - it is needed to find eigen in the linux docker image)
5. A Dockerfile for the Linux wheel builder
6. A patch to auditwheel until the associated PR gets merged
7. An added template instantiation to electronthresholds.cpp to avoid an import error in the python wheels

Mac and Windows wheels should hopefully use a lot of the same code paths and require just a little bit of editing...


CMakeLists.txt
```diff
@@ -52,11 +49,20 @@ set(_src
...
-  stempy/electronthresholds.cpp
-  h5cpp/h5readwrite.cpp)
+  stempy/electronthresholds.cpp)
 
 set(_libs Threads::Threads)
-list(APPEND _libs "${HDF5_LIBRARIES}")
+list(APPEND _libs pybind11::headers)
+# We have to link to Python3::Module since "pyreader.cpp" includes
+# pybind11 and python header files.
+list(APPEND _libs Python3::Module)
```
cjh1, line 58: So the `stempy` library now depends directly on Python?

scripts/github-actions/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
+ARG BASE_IMAGE=quay.io/pypa/manylinux2010_x86_64
```
cjh1, line 1: May be this should live under `docker/github-actions/`?

stempy/reader.h
```diff
@@ -569,7 +582,7 @@ std::future<void> SectorStreamMultiPassThreadedReader::readAll(Functor& func)
   }
 
   // Reset counter
-  m_processed = { 0 };
+  m_processed = 0;
```
cjh1, line 585: Why this change?

psavery: It fixed lots of warnings on Mac. Hopefully it will work at runtime...

CMakeLists.txt
psavery: Yes, just because `pyreader.cpp` is compiled as part of the source, and `pyreader.h` files are included.

scripts/github-actions/docker/Dockerfile
psavery: We tried moving `scripts/github-actions` to `.github/scripts`. Let's hope it works! :crossed_fingers:

cjh1 (approved):
LGTM, Thanks, lets get it merge if the CI is successful.

PR #197, opened 2021-03-02
Skip pip install stempy for read the docs

psavery:
It isn't really needed - we are already appending the path
to the python files to the sys.path. And we are mocking the
stempy._io and stempy._image imports already.

This also adds _threaded_multi_pass_reader to the list of _io_mock
objects to prevent an error involving multiple inheritance with
different metaclasses.

deprecation is also added as a requirement for the docs, since it
is needed to display the deprecation in the documentation.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,6 +6,4 @@ sphinx:
 python:
   version: 3.7
   install:
-    - method: pip
-      path: .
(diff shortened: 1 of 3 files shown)


cjh1 (approved):
LGTM, thanks for fixing this up.

PR #199, opened 2021-03-05
Phase from COM

ercius:
Add function to calculate the phase from the center of mass.  This is mostly copied from py4DSTEM and was authored by @HamishGBrown


cjh1 (approved):
LGTM

Issue #201, opened 2021-03-08
Jupyter docker image issue


ercius:
I updated the stempy jupyter image just now. It looks like stemy.io and stempy.image can't be imported. I can import stempy.  I am able to load stempy.io and stempy.image in the non-jupyter version.

I will revert to the older version so this is not critical. But it would be good to look at soon.

cjh1:
I think this is related to the new wheels, sorry about that. @psavery is going to take a look.

psavery:
Hi @ercius. Do you mean the [stempy-scipy-notebook image](https://hub.docker.com/r/openchemistry/stempy-scipy-notebook)? I was able to start up a jupyter notebook with it and load the modules.

![Screenshot from 2021-03-08 15-38-32](https://user-images.githubusercontent.com/9558430/110385936-366a1800-8025-11eb-90b5-c42ab9bef59f.png)

Let me know if you are doing something different.

cjh1:
@ercius that would correspond to the following image `openchemistry/stempy-scipy-notebook` you would have pulled.

ercius:
@cjh1 I pulled that image. On mothership6 Im running:
`singularity exec -B /mnt:/mnt,/tmp:/home/jovyan ~/stempy-scipy-notebook_latest.sif jupyter lab`
Then I connect to the Jupyter Lab server. 

Then `import stempy` works but `import stempy.io as stio` and `from stempy import io` do not work.

I just re-pulled the image and got a lot of changes. However, I still can not import `stempy.io`.

Here is some information about the image:
```bash
>> singularity inspect stempy-scipy-notebook_latest.sif
org.label-schema.build-date: Wednesday_10_March_2021_9:20:27_PST
org.label-schema.schema-version: 1.0
org.label-schema.usage.singularity.deffile.bootstrap: docker
org.label-schema.usage.singularity.deffile.from: openchemistry/stempy-scipy-notebook:latest
org.label-schema.usage.singularity.version: 3.6.1+11-g7c3204ca4
```

psavery:
@ercius I see, I was able to reproduce this in singularity. I guess this might be a specific problem to singularity. I'll look into it.

cjh1:
@ercius This should now be resolved, please try pulling the latest `openchemistry/stempy-scipy-notebook` image.

ercius:
Fixed! Thanks!

PR #202, opened 2021-03-11
Dereference symlinks to fix singularity

psavery:
Previously, symlinks were being preserved when the copy occurred. The
symlinks pointed into the /home/jovyan directory.

When a singularity shell was started, the user would be "jovyan", so
there would be no issue importing the files.

However, if jupyter was started instead, the user would be the current
user on the host OS. And this user would have permissions issues with
reading into the /home/jovyan directory.

To fix this, dereference the symlinks on `cp` to avoid having symlinks
pointing into `/home/jovyan`.

This was tested and seems to have fixed the issue.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -82,7 +82,7 @@ RUN mkdir -p /build/stempy && \
 
 # Install stempy
 RUN pip3 install -r /source/stempy/requirements.txt && \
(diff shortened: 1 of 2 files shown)


cjh1 (approved):
LGTM

cjh1:
Fixes #201

PR #206, opened 2021-04-15
fix error in header description. 

ercius:
The frame number was described as unit16 when it is actually uint32. The numpy custom dtype is correct though.


ercius:
Also, can we add this document into the read the docs table of contents? Im not sure how to do that.

PR #218, opened 2021-11-01
Skip musl wheels

psavery:
~It's end of life is coming in \~7 weeks, and the latest cibuildwheel
no longer supports it.~

~We can go ahead and remove it, since we do not believe anyone has a
specific need for the python 3.6 wheels.~

Looks like it was actually musl wheels that were the problem. CI build wheel started supporting musl wheels since a couple of weeks ago. We don't need any immediate support for them, so I added them to the skip list, and I put python 3.6 back on.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -9,9 +9,9 @@ on:
   pull_request:
 
 env:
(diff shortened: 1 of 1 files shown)


cjh1 (approved):

cjh1 (commented):
Great! Thanks

Issue #210, opened 2021-09-17
synchronize names in electron_counts and SparseArray


ercius:
`stempy.io.load_electron_counts` returns a named tuple with `frame_dimensions` and `scan_dimensions`

`SparseArray.from_hdf5` returns a class with `frame_shape` and `scan_shape`

These should probably be the same. I think using shape is better because it matches `ndarray.shape`

PR #222, opened 2021-12-17
Replace ElectronCountedData with SparseArray

psavery:
This removes all occurrences of the `ElectronCountedData` namedtuple and
replaces them with a `SparseArray` object. However, the `create_stem_images()`
method still accepts `ElectronCountedData` namedtuple arguments in case
any are still in use from older versions of stempy.

Fixes: #210


psavery:
I've tested a few examples, and I added a round trip test for `SparseArray`'s HDF5 I/O. But I haven't tested everything, partly because I don't have the data needed for all of the examples.

cjh1:
@psavery So this is going to be a breaking change? As the return type has changed? Or is `SpareArray` compatibile with `ElectronCountedData`?

psavery:
It is a breaking change. The `SparseArray` has similar attributes as the `ElectronCountedData`, but `scan_dimensions` is now `scan_shape[::-1]`, and `frame_dimensions` is now `frame_shape`.

Here are the breaking changes:

- `stempy.image.electron_count()` now returns a `SparseArray` instead of the namedtuple
- `stempy.io.load_electron_counts()` now also returns a `SparseArray` 
- `stempy.io.save_electron_counts()` now requires a `SparseArray` argument instead of the namedtuple

Note that `stempy.image.create_stem_images()` can still accept an `ElectronCountedData` namedtuple as input, so that should still work.

cjh1:
OK, we will have to bump the major version. @ercius you will need to check if this is going to effect any of your processing script.

ercius:
I think the Distiller counting script is the most important to check. It uses the output of `stempy.image.electron_count()` to save to HDF5.

Otherwise, I have moved most of my scripts over to SparseArray already.

ercius:
Also, can we version the Docker images? Then its possible to use the current `stempy` version.

cjh1:
> I think the Distiller counting script is the most important to check. It uses the output of `stempy.image.electron_count()` to save to HDF5.

The distiller script should be fine as it just passes the result of `electron_count` to `save_electron_count`

ercius:
The best way for me to test this is with a new release. I tried to make a conda environment and `pip install -e` to develop, but I got an error. Ill post that when I get a chance.

cjh1:
> Also, can we version the Docker images? Then its possible to use the current `stempy` version.

We can push tags, we currently just build `latest` ( that tracks master ). I will push a tag for the latest version. However, you will need to make sure you pull that specific version. Is this ms6 you are thinking of?

cjh1:
> The best way for me to test this is with a new release. I tried to make a conda environment and `pip install -e` to develop, but I got an error. Ill post that when I get a chance.

Yes, not sure we have put any time into getting the dev environment working. When this is merged a "pre-release" package will be pushed to pypi, that may be the easiest what for you to test it. Each push also builds wheels on github actions, so you have down load the "artifact" and install locally as well. For example [here](https://github.com/OpenChemistry/stempy/suites/4696063136/artifacts/128298398) are the wheels for that PR

ercius:
> We can push tags, we currently just build latest ( that tracks master ). I will push a tag for the latest version. However, you will need to make sure you pull that specific version. Is this ms6 you are thinking of?

I think it would be good to have the 1.1.1 version in case I need to revert on MS6 or Jupyter@NERSC. Some users are using copies of my analysis scripts, which will be harder to update. So, being able to go revert would be ideal just in case.

ercius:
> Yes, not sure we have put any time into getting the dev environment working. When this is merged a "pre-release" package will be pushed to pypi, that may be the easiest what for you to test it. Each push also builds wheels on github actions, so you have down load the "artifact" and install locally as well. For example here are the wheels for that PR

No problem then. I dont think a pre-release is necessary. Ill be able to adapt any code pretty easily on the fly. Its not a huge breaking change. Ill check out the "artifact" wheels as a way to test versions.

psavery:
@ercius In my commit I just pushed, I removed a bunch of old examples that appear to be using old data formats, or appear that we haven't touched them in a long time. Let me know if I removed anything you want to bring back.

cjh1:
> I think it would be good to have the 1.1.1 version in case I need to revert on MS6 or Jupyter@NERSC. Some users are using copies of my analysis scripts, which will be harder to update. So, being able to go revert would be ideal just in case.

OK, I have pushed 1.1.1 tags for our docker images in preparation. Note that for Jupyter@NERSC you would need to adjust your `~/.local/share/jupyter/kernels/stempy-jupyterlab/kernel.json` if you want to fix to 1.1.1.

ercius:
@psavery I confirm removing all of the example files this PR removes.

ercius:
Thanks! That's will allow us to go back if there are issues. I don't expect
any though.

psavery:
@cjh1 This is ready for review.

cjh1 (approved):
LGTM Thank you!

PR #223, opened 2021-12-30
Add extra for loop to account for multiple frames

ercius:
This fixes the issue with `stempy.image.calculate_sum_sparse()` which incorrectly calculated results for scans with multiple frames per probe position


cjh1 (approved):
LGTM

PR #225, opened 2022-01-13
Remove deprecated functions flagged for 1.1

psavery:
These functions were flagged for deprecation starting in 1.1, which
we are now on.


cjh1 (approved):
LGTM

PR #226, opened 2022-01-13
Add metadata to SparseArray

psavery:
This adds metadata to the SparseArray class than will be saved/loaded
via HDF5. A basic test was also added to ensure this works.

Now, when electron counting is performed, the metadata on the sparse
array gets set to the electron counting metadata (thresholds mostly).
This way, we can also keep and save the electron counting metadata to
the HDF5 file.

Fixes: #220

### DIFF
--- a/python/stempy/image/__init__.py
+++ b/python/stempy/image/__init__.py
@@ -410,11 +410,13 @@ def electron_count(reader, darkreference=None, number_of_samples=40,
 
     # Convert to numpy array
     np_data = np.array([np.array(x, copy=False) for x in data.data], dtype=np.object)
+    metadata = _electron_counted_metadata_to_dict(data.metadata)
 
     kwargs = {
         'data': np_data,
         'scan_shape': data.scan_dimensions[::-1],
         'frame_shape': data.frame_dimensions,
+        'metadata': {'electron_counting': metadata},
     }
     array = SparseArray(**kwargs)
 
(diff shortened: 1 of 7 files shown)


psavery:
I still need to test the electron counting part before this PR is ready. I'll do so when I have some data to test it on.

python/stempy/image/__init__.py
```diff
@@ -415,6 +415,7 @@ def electron_count(reader, darkreference=None, number_of_samples=40,
         'data': np_data,
         'scan_shape': data.scan_dimensions[::-1],
         'frame_shape': data.frame_dimensions,
+        'metadata': {'electron_counting': data.metadata},
```
psavery, line 418: This probably won't work, but I'll fix it when I get to try with real data.

psavery:
This is now working, and I modified the `electron_count.py` example to use the metadata instead. You can see some sample output from `h5dump -g /metadata` [here](https://github.com/OpenChemistry/stempy/files/7866481/metadata_output_example.txt).

ercius:
@psavery It looks like you used datasets for all of the metadata. Would it be better to use attributes? Or is it more typical to use datasets for this type of information?

cjh1:
We could probably use attributes attached to groups. This would still have the flexibility to have a hierarchy rather than a flat structure just applied only to the dataset.

psavery:
@cjh1 I was thinking the same thing

psavery:
@ercius I've updated it to store the metadata values in attributes instead. It still uses groups to preserve the hierarchy. An example of `h5dump -g /metadata` output is [here](https://github.com/OpenChemistry/stempy/files/7885263/out.txt).

cjh1 (approved):
LGTM

PR #227, opened 2022-01-13
Warn if electron_count() is called with old reader

psavery:
These old data formats mostly do not even exist anymore. We will remove
them in the future.


cjh1 (approved):
LGTM

Issue #208, opened 2021-09-02
Modify sparse data result for scans with multiple positions


ercius:
we have the ability to acquire multiple frames per scan position and stempy will count that data. It currently combines all events from the same scan position into one event list in the final sparse data set. See #176. 

I would like to change this behavior so that each frame is a separate list and is associated with a scan position. This capability is builtin to the sparse data format but not user. See description [here](https://stempy.readthedocs.io/en/latest/StempyH5Format.html).

If we take 1 frame per scan position then `len(frames) == n*m` where the scan is nxm number of scan positions. `scan_positions` is then a list of `[0,1,2,3,...,n*m]`

If we take 2 then `len(frames) == 2*n*m` and `scan_positions` should be a list like `[0,0,1,1,2,2,3,...,n*m,n*m]`

To avoid breaking older analysis code I think we might want to indicate the new sparse data is different somehow. Or maybe provide the option to save the sparse data set in the old way (combining all frames into one event list) and the new way described above.

psavery:
@ercius, could you provide me with a test dataset?

ercius:
Yes. Try this one:
distiller scan id 505
https://distiller.lbl.gov/scans/505

the counted data is on NERSC in the ncemhub/distiller/distiller/counted/2022.01.04/data_scan5_...

There are 2 frames per position for that scan.

psavery:
@ercius Is the raw data available for this? I'll need that to test changes to the electron counting methods in stempy.

ercius:
no...that is deleted. Ill have to take a new data set to be sure we have exactly what you need. Ill take a small data set soon, store the raw data on NERSC, and send you the location.

psavery:
Sounds great, thank you Peter!

psavery:
@ercius I'd also like to confirm that I'm giving the `SparseArray` operations the correct behavior for scans with multiple frames per scan position.

I'm going to make up an example. Let's say we have two scan positions and each position has two 2x2 frames. At scan position 0, the sparse arrays are `[0]` and `[0, 2]`, and at scan position 1, the sparse arrays are `[0]` and `[0, 1]`. Can you confirm if all of these are correct?

1. The expanded dense arrays for the two scan positions would be:
```python
# position 0
[[2, 0],
 [1, 0]]

# position 1
[[2, 1],
 [0, 0]]
```

2. The sum would be:
```python
[[4, 1],
 [1, 0]]
```

3. The max would be:
```python
[[2, 1],
[...]

PR #228, opened 2022-01-17
Separate sparse arrays with same scan position

psavery:
Previously, in the electron counting, we would combine two sparse arrays
if they had the same scan positions. Now, we are modifying this so that
the sparse arrays are kept separately per frame, and duplicates are no
longer allowed in a single sparse array.

The ElectronCountedData struct now has a scanPositions vector to indicate
which scan a sparse array belongs to, since the data vector no longer
matches the size of the scan.

This also updates the SparseArray class to perform faster (and correct)
arithmetic and expanding based upon the new way that the sparse data is
formatted.

I have tested the electron counting with a dataset that has two frames
per scan position, and for fully expanding the output sparse array, I
received the same result before and after these changes, which indicates
that it is hopefully working correctly.

An extensive test was also added which tests most pieces of using a
SparseArray where some scan positions have multiple frames.

This will probably have some minor merge conflicts with #226.

Fixes: #208

### DIFF
--- a/python/image.cpp
+++ b/python/image.cpp
@@ -38,11 +38,13 @@ struct ElectronCountedDataPyArray
       data.push_back(vectorToPyArray(std::move(vec)));
     }
 
+    scanPositions = other.scanPositions;
     scanDimensions = other.scanDimensions;
     frameDimensions = other.frameDimensions;
   }
 
   std::vector<py::array_t<uint32_t>> data;
+  std::vector<uint32_t> scanPositions;
 
   Dimensions2D scanDimensions = { 0, 0 };
   Dimensions2D frameDimensions = { 0, 0 };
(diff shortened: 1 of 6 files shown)


cjh1:
@psavery I have resolved the merge conflicts, please check I did the right think :-) I just had one question what changes a user would need to make to their code to adapt to these changes ( for the release notes )?

psavery:
I think for this PR, no code changes are needed in most cases.

The main changes are how the electron counted data is stored for multiple frames per scan. Rather than having duplicate values in a single array, there can be more than one array per scan position. But this is a low-level detail, and the changes are handled automatically by the `SparseArray`.

If someone was using the data directly, though, instead of through the `SparseArray`, and they were working with multiple frames per scan, they'd have to modify their code to check the `scan_positions` to see which scan position each frame belongs to, and to recognize that there may be more frames than there are scan positions.

cjh1 (approved):

PR #230, opened 2022-03-02
Modify MPI to work with new multi-frame support

cjh1:


cjh1:
@psavery I think this fixes up the MPI based code path.

psavery (approved):
LGTM

Issue #239, opened 2022-03-09
com_sparse does not work with bin_scans


ercius:
`SparseArray.bin_scans()` does not change the `SparseArray.data` element. This means that `com_sparse` is unable to work with data that is binned in this way. The algorithm needs to be changed to since center of mass of each pattern is not separable.

ercius:
- [ ] Change algorithm to work with bin_scans()
- [ ] Test it works with multiple scans per position
- [ ] update input to allow SparseArray() without needing extra keywords

psavery:
@ercius Can I update it to take just a sparse array and the `crop_to` argument, or do you want to preserve the other arguments as options?

psavery:
(Note, if we preserve the other arguments, we will most likely need to add a `scan_positions` argument as well to take into account the scan positions now).

ercius:
I think that we can just use the SparseArray on second thought. So, you can remove the other arguments.

What will happen to older data in the old format? Will that still work or do we need to use 1.1.0?

psavery:
So, I think data in the older format will still work as long as there are not multiple frames per scan position.

However, if we do have a bunch of data in the older format with multiple frames per scan position, I'm happy to also write some code to convert it to the new format. It could also be relatively easy to detect as well - we can check to see if there are any duplicate values in any of the sparse frames (which would imply it's using the older format), and then convert it to the new format.

If there's only one frame per scan position, then the new format and the old format are identical.

psavery:
Hi @ercius. I think I have this working, along with a test. But just to confirm, I wanted to provide you my example and the results. It should match the results of using the old format with the previous version of the function.

For a SparseArray of `scan_shape == (2, 1)` and `frame_shape == (2, 2)`, and `scan_positions == (0, 0, 1, 1)`, where the two sparse frames at scan position 0 are `[0]` and `[3]`, and the two sparse frames at scan position 1 are `[0]` and `[0, 1, 2]`, the expanded frames would look like these (in order):
```python
# scan position 0
[[1, 0],
 [0, 0]]

[[0, 0],
 [0, 1]]

# scan position 1
[[1, 0],
 [0, 0]]

[[1, 1],
 [1, 0]]
```

And the COM for each scan position would be this:
```python
[[ 0.5], [ 0.5]] # scan position 0
[[0.25], [0.25]] # scan position 1
```

ercius:
[...]

PR #241, opened 2022-03-10
Update com_sparse() for new data format

psavery:
This now uses the `SparseArray` class, and it takes into account the
fact that there may be multiple frames per scan position.

Fixes: #239

### DIFF
--- a/python/stempy/image/__init__.py
+++ b/python/stempy/image/__init__.py
@@ -406,7 +406,9 @@ def com_sparse(array, crop_to=None):
             ev = []
 
         if len(ev) > 0:
-            x, y = np.unravel_index(ev, array.frame_shape)
+            x = ev // array.frame_shape[0]
+            y = ev % array.frame_shape[1]
+
             mm0 = len(ev)
             comx0 = np.sum(x) / mm0
             comy0 = np.sum(y) / mm0
--- a/python/stempy/image/__init__.py
+++ b/python/stempy/image/__init__.py
@@ -380,19 +380,13 @@ def maximum_diffraction_pattern(reader, darkreference=None):
     return img
 
 
-def com_sparse(electron_counts, frame_dimensions, scan_dimensions, crop_to=None):
+def com_sparse(array, crop_to=None):
     """Compute center of mass (COM) for counted data directly from sparse (single)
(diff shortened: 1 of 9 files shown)


python/stempy/image/__init__.py
```diff
@@ -402,10 +396,17 @@ def com_sparse(electron_counts, frame_dimensions, scan_dimensions, crop_to=None)
...
+        data_indices = array.data_indices(scan_position)
+        if len(data_indices) != 0:
+            # Combine the sparse arrays into one array that we will use
+            ev = np.hstack(array.data[data_indices])
+        else:
+            ev = []
+
         if len(ev) > 0:
-            x, y = np.unravel_index(ev, frame_dimensions)
+            x, y = np.unravel_index(ev, array.frame_shape)
```
ercius, line 409: I have found that `np.unravel_index()` is slower than simply:
```python
x  = ev // frame_shape[0]
y = ev % frame_shape[1]
```
For speed, we might be consider using this instead to avoid the overhead of calling the function and all the extra checks it does.

psavery: This is now fixed.

tests/requirements.txt
```diff
@@ -1,2 +1,3 @@
 pytest
 requests
+scipy
```
psavery, line 3: Note that `scipy` was added as a test dependency so we can compare the `com_sparse()` results with `scipy.ndimage.center_of_mass()` results.

Strangely, it seems that adding the `scipy` dependency added about 15 minutes to the Ubuntu CI time, because pip for python3.8 and python3.9 is having trouble finding/building a compatible scipy dependency for the tests. It eventually downgrades scipy a few times and finds one that works, but a lot of time is being spent trying to build newer versions of scipy, which fails. I don't know why it can't just install some of the `manylinux` scipy wheels that are already on pypi.

But, the tests still pass, even though it takes extra time.

ercius: stempy also has stempy.image.com_dense(). That would be a lot easier with no need for new dependencies.

psavery: Thanks Peter! I didn't realize we had that one. I just switched it out.

cjh1 (approved):
LGTM

Issue #238, opened 2022-03-09
add version


ercius:
add `stempy.__version__` (or similar) for convenience of packaging.

PR #244, opened 2022-03-16
Add stempy version

psavery:
This can be obtained via `stempy.__version__`.

This also modifies some cmake logic to remove the need to specify every python file and directory for installation.

Fixes: #238


cjh1 (approved):
LGTM

Issue #243, opened 2022-03-15
SparseArray and other image functions


ercius:
create_stem_images crashes the Jupyter kernel for SparseArray input with multiple probes per position. We should check that all functions in 2.0 and make sure they are compatible with the new file structure.

cjh1:
@psavery Can you take a look at this?

psavery:
I think I see what's going on. @cjh1 let's chat about this when we get a chance.

psavery:
@ercius I was able to reproduce the crash and fix it in #245. I confirmed as well that I obtained the correct result for my example test.

PR #245, opened 2022-03-16
Fix create_stem_images() crash with v2 data

psavery:
This fixes a crash that would occur if `create_stem_images()` was
called using v2 data, and there were more than one frame per scan
position.

Internally, it does so by moving around an additional array of scan
positions to keep track of which frame belongs to which scan position.

The python API remains unchanged.

Fixes: #243


cjh1 (approved):
LGTM

PR #248, opened 2022-03-31
Fix sparse slicing performance issues

psavery:
Too many large arrays of booleans were being created and slowing down
the sparse slicing significantly for large scan sizes.

This was particularly noticeable, for instance, for a scan shape of
(1024, 1025), where we would crop off one of the scans like

```
array[:, :1024]
```

These changes reduce the number of large boolean arrays created and
seem to significantly improve the performance.

These changes will also be unnecessary and removed in V3 of the
SparseArray, which is coming soon.


cjh1:
@psavery Seems like the building of the wheels if failing?

cjh1 (approved):
Nice work!

PR #249, opened 2022-04-01
Refactor SparseArray and methods to data version 3

psavery:
Version 3 of the SparseArray and electron counting methods changes the
data shape from a flat array of sparse frames to a 2D array, where the
first dimension is the scan position, and the second dimension is the
index of a frame at that scan position. This removes the need for the
`scan_positions` array (previously used to indicate which scan position
each sparse frame corresponded to).

Now, the data array passed to the SparseArray must either have a shape
of `(num_scans, frames_per_scan)` or just `(num_scans,)` (where one frame
per scan will be assumed, and the shape will be changed automatically to
`(num_scans, 1)`. Validation is performed to ensure this is the case, or
an exception will be raised.

This simplifies a lot of the logic involving multiple frames per scan
position, and completely removes the need for manipulating the scan
positions array that the SparseArray previously held.

For writing to HDF5, it appears that h5py does not allow a 2D array of
variable length arrays to be written (see h5py/h5py#876), so we must
flatten it to a 1D array for HDF5 storage. The scan shape allows the
2D array to be re-created on loading.

Backwards compatibility has been maintained so that the SparseArray class
can load in HDF5 files saved in versions 1, 2, or 3 and correctly convert
them to version 3 upon loading. Tests have been added to ensure this is
the case.

### DIFF
--- a/python/stempy/io/sparse_array.py
+++ b/python/stempy/io/sparse_array.py
@@ -80,17 +80,18 @@ class SparseArray:
     and stride. Slicing returns either a new sparse array or a dense
     array depending on the user-defined settings.
     """
-    VERSION = 2
+    VERSION = 3
 
     def __init__(self, data, scan_shape, frame_shape, dtype=np.uint32,
-                 sparse_slicing=True, allow_full_expand=False,
-                 scan_positions=None, metadata=None):
+                 sparse_slicing=True, allow_full_expand=False, metadata=None):
         """Initialize a sparse array.
 
-        :param data: the sparse array data, where the arrays represent
-                     individual frames, and each value in the arrays
+        :param data: the sparse array data, where the outer array represents
+                     scan position, the middle array represents a frame taken
+                     at the scan position, and the inner array represents the
+                     sparse frame data, where each value in the arrays
                      represent 1 at that index.
-        :type data: np.ndarray (1D) of np.ndarray (1D)
+        :type data: np.ndarray (2D) of np.ndarray (1D)
         :param scan_shape: the shape of the scan
         :type scan_shape: tuple of length 1 or 2
         :param frame_shape: the shape of the frame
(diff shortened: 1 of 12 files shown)


stempy/electron_mpi.cpp
```diff
@@ -88,8 +87,7 @@ void sendPartialMap(std::vector<Events>& events)
            MPI_BYTE, 0, 0, MPI_COMM_WORLD);
 }
 
-void gatherEvents(int worldSize, int rank,
-                 std::vector<Events>& events)
+void gatherEvents(int worldSize, int rank, Events& events)
```
psavery, line 90: @cjh1 For the MPI functions, I noticed that a `std::vector<Events>&` was being used, where the outer vector represented the scan position.

Now, an extra vector was added to the `Events` object, and the outer vector represents the scan position for it. So I am hoping that just removing the `std::vector<>` from around the `Events` will be enough. It compiles at least. Can you try it out and see if it works?

python/stempy/io/sparse_array.py
```diff
@@ -759,40 +798,10 @@ def slice_range(slice, length):
...
-            # Re-number the positions so they go from zero to
-            # len(positions_to_keep).
-            # Sort them so we don't undo the new ordering which is already
-            # a part of the new_frames.
-            unique = np.unique(np.sort(new_scan_positions))
-
-            # Create a map of old values to new values
-            conversion_map = {x: i for i, x in enumerate(unique)}
-            for i, pos in enumerate(new_scan_positions):
-                new_scan_positions[i] = conversion_map[pos]
```
psavery, line 791: @cjh1 V3 removed a bunch of code in this function...

psavery:
@cjh1 I left a comment about the MPI changes. Can you check on these two things?

- [ ] Check that MPI electron counting works
- [ ] Check what happens to data that produced no sparse frames

cjh1 (approved):
LGTM

PR #253, opened 2022-04-22
Radial sum sparse v3

### LINKED ISSUE #240: Allow SparseArray input to radial_sum_sparse input
use SparseArray to automatically fill scan_shape and frame_shape

leave those as keywords for manual use cases and older code.

ercius:
This should fix #240 for v3 data.

I also wrote a pytest function and a new fixture for this as well. I was not able to test that it actually runs though because I dont have v3 compiled properly. Can you pull this and check that my test passes?


python/stempy/image/__init__.py
```diff
@@ -500,35 +500,31 @@ def calculate_sum_sparse(electron_counts, frame_dimensions):
...
+
+    for ii, frames in enumerate(sa.data):
+        # Loop over all scan positions
+        for jj, events in enumerate(frames):
+            # Loop over all frames in this scan position
+            x, y = np.unravel_index(events, sa.frame_shape)
+            r = np.sqrt((x - center[0]) ** 2 + (y - center[1]) ** 2)
+            hh, hhx = np.histogram(r, bins=range(0, num_bins + 1))
+            r_sum[ii, :] += hh
+    r_sum = r_sum.reshape((sa.scan_shape[0], sa.scan_shape[1], num_bins))
```
psavery, line 527: Note that because the previous `scan_dimensions` is equal to the new `scan_shape[::-1]`, maybe this should be `sa.scan_shape[1]`, `sa.scan_shape[0]`?

tests/test_image.py
```diff
@@ -35,3 +35,10 @@ def test_com_sparse(sparse_array_small, full_array_small):
             full_com[:, i, j] = com_dense(full_array_small[i, j])[:, 0]
 
     assert np.array_equal(com, full_com)
+
+
+def test_radial_sum_sparse(sparse_array_10x10):
+
+    rr = radial_sum_sparse(sparse_array_10x10, center=(4, 4))
+
+    assert np.all(rr[0, 0, :] == [0, 6, 0, 0, 3])
```
psavery, line 44: It appears that this is failing in the tests, where `rr[0, 0, :]` is actually coming out to be `[3, 0, 3, 0, 0]` instead. @ercius any idea of what might be wrong?

Note that using `np.array_equal()` might be a little faster as well: `np.array_equal(rr[0, 0, :], [0, 6, 0, 0, 3])`

python/stempy/image/__init__.py
ercius: I dont think so. The radial sum is performed on the frames. The SparseArray that is returned should have the same scan_shape as the input.

ercius:
Another issue that I want to make sure we resolve is the ordering of the `center` input. Does this match what we expect for the ordering of the frame dimensions? Or should I swap the order of `center` in this line:
`r = np.sqrt((x - center[0]) ** 2 + (y - center[1]) ** 2)`

I made a test for this which I hope makes sense. I first crop axis 2 for a SparseArray to make it non-square in the frame_shape. There is a disk centered at about 1/2 of axis 2 and axis 3. I can then set the center to approximately the center of each axis but in different orders and test the output. In the case of `center=(300, 136)` the first number is actually larger than the axis shape and there cannot be a disk there. The proper output should show a linear increase in intensity starting at radius = 0 as seen by the magenta curve generated using `center=(136,300)`. It looks right, but can verify my logic?
```python
sa.allow_full_expand = True
sa_dp_crop = sa[:,:,0:288,:] # crop axis 2 to make non-square

fg, ax= plt.subplots(1, 2)
ax[0].imshow(sa_dp_crop.sum(axis=(0,1)))
r_axis3_2 = radial_sum_sparse(sa_dp_crop, (300, 136))
r_axis2_3 = radial_sum_sparse(sa_dp_crop, (136, 300)) # should be centered on a disk
ax[1].plot(r_axis3_2.sum(axis=(0, 1)),c='r',label='no disk')
ax[1].plot(r_axis2_3.sum(axis=(0, 1)),c='m',label='disk')
ax[1].legend()
```
Figure 49

tests/test_image.py
ercius: My test is incorrect. The line should be:
`rr = radial_sum_sparse(sparse_array_10x10, center=(5, 5))`
and then the output `[0, 6, 0, 0, 3]` is correct.

Ill also update to using `np.array_equal()`

psavery:
> Another issue that I want to make sure we resolve is the ordering of the `center` input. Does this match what we expect for the ordering of the frame dimensions? Or should I swap the order of `center` in this line: `r = np.sqrt((x - center[0]) ** 2 + (y - center[1]) ** 2)`
> 
> I made a test for this which I hope makes sense. I first crop axis 2 for a SparseArray to make it non-square in the frame_shape. There is a disk centered at about 1/2 of axis 2 and axis 3. I can then set the center to approximately the center of each axis but in different orders and test the output. In the case of `center=(300, 136)` the first number is actually larger than the axis shape and there cannot be a disk there. The proper output should show a linear increase in intensity starting at radius = 0 as seen by the magenta curve generated using `center=(136,300)`. It looks right, but can verify my logic?
> 
> ```python
> sa.allow_full_expand = True
> sa_dp_crop = sa[:,:,0:288,:] # crop axis 2 to make non-square
> 
> fg, ax= plt.subplots(1, 2)
> ax[0].imshow(sa_dp_crop.sum(axis=(0,1)))
> r_axis3_2 = radial_sum_sparse(sa_dp_crop, (300, 136))
> r_axis2_3 = radial_sum_sparse(sa_dp_crop, (136, 300)) # should be centered on a disk
> ax[1].plot(r_axis3_2.sum(axis=(0, 1)),c='r',label='no disk')
> ax[1].plot(r_axis2_3.sum(axis=(0, 1)),c='m',label='disk')
> ax[1].legend()
> ```

I think it looks good to me! Also, we did not change the ordering of the older `frame_dimensions` and the new `SparseArray.frame_shape`, so if the previous ordering was correct, then it should be good to go.

cjh1 (approved):

PR #254, opened 2022-04-22
Notebooks v3

ercius:
These example notebooks should work with v3 data. They also have embedded output so they render on Github.


psavery:
@ercius The examples look really nice!

One question I had is that I see code like this in both of the files:
```python
# Calculate number of electrons per pattern
num_elec = 0
for ii in vacuum_scan.data:
    num_elec += len(ii)
avg_elec = num_elec / vacuum_scan.data.shape[0]
```

Do we need to change this for v3, to maybe be something like this?

```python
# Calculate number of electrons per pattern
num_elec = 0
for ii in vacuum_scan.data:
    for jj in ii:
        num_elec += len(jj)
avg_elec = num_elec / vacuum_scan.data.shape[0]
```

Note the extra loop since the data is now 2D, where you have to loop twice to reach the actual sparse frame. This would give the average number of electrons per scan (note: which is *not* the same as average number of electrons per frame).

ercius:
You are correct about that example. Ill add that second loop in and update it.

ercius:
Let me know if you have any more comments on the new changes.

psavery:
@ercius Looks good to me!

ercius:
@psavery @cjh1 I think the PR needs to be approved by one of you first.

PR #256, opened 2022-07-10
Add two new options to com_sparse. 

ercius:
Allow user to keep nans for empty frames. This does not fix all missing data problems like missing sectors which affect the com.

Allow user to input initial center but default to using the center of mass of the entire frame.

### DIFF
--- a/python/stempy/image/__init__.py
+++ b/python/stempy/image/__init__.py
@@ -392,17 +392,21 @@ def maximum_diffraction_pattern(reader, darkreference=None):
     return img
 
 
-def com_sparse(array, crop_to=None):
+def com_sparse2(array, crop_to=None, init_center=None, replace_nans=True):
     """Compute center of mass (COM) for counted data directly from sparse (single)
         electron data. Empty frames will have the average COM value of all frames. There is an option to crop to a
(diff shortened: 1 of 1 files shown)


ercius:
I added a new function as well. It might work in some cases to filter out bad scan positions. I need to see how it works in a variety of cases.

python/stempy/image/__init__.py
```diff
@@ -392,17 +392,21 @@ def maximum_diffraction_pattern(reader, darkreference=None):
     return img
 
 
-def com_sparse(array, crop_to=None):
+def com_sparse2(array, crop_to=None, init_center=None, replace_nans=True):
```
cjh1, line 395: I don't think we need to rename? The defaults for the new parameters should give the old behavior?

ercius: That is a mistake. I was testing it in another file and forgot to change it back. I just pushed a fix.

cjh1 (approved):

cjh1:
@ercius Cutting a new release and will rebuild images ...

cjh1:
@ercius I have built new jupyter images, they are available on cori.

Issue #258, opened 2022-07-25
Add row-dark operation to counting


ercius:
Each frame has a global dark (pedestal) subtracted during acquisition. There is a secondary pre-frame dark that can be applied which improves the counting quality. The detector is divided into 2 halves and a series of rows. Here is a quick diagram:
11111 | 11111
22222 | 22222
33333 | 33333
44444 | 44444

A row dark is calculated as the mean (or median) of each row on each side of the split in the center. Here is python code:
```python
# Determine the mean of each row for each side (detector is split down the middle)
raws_avg0 = np.mean(frame[:,:],axis=1) # might want to use median
raws_avg1 = np.mean(frame[:,:],axis=1)
            
# The row-dark is the mean of each row
dark[:,:288] = raws_avg0[:,None]
dark[:,288:] = raws_avg1[:,None]
            
# Divide out the row-dark and multiply by the data set's mean
# i.e. ****Statistics for calculating electron thresholds****
#      optimized mean: 19.477028281405794
f1['frames'][ii, :, :] = optimized_mean * frame / dark
```
This should be calculated and applied to each frame separately.

psavery:
Hi @ercius. For the example above, the `raw_avg0` and `raw_avg1` appear to be identical. Is that supposed to be the case?

ercius:
Good question. That was not intended. I mostly meant to label rows with their row number rather than putting a value in that row. A better labeling would be (L = left and R = right)
1L 1L 1L 1L 1L | 1R 1R 1R 1R 1R
etc.

And a better labeling of the possible pixel values is:
20 20 20 19 20 | 21 20 20 21 21
etc.

1L's mean is 24.75 and 1R's mean is 26.5. So the row dark is
24.75 24.75 24.75 24.75 24.75 | 26.5 26.5 26.5 26.5 26.5 
etc.

psavery:
Thanks @ercius! However, I meant the code example actually. It appears that `raws_avg0` and `raws_avg1` are both equal to `np.mean(frame[:,:],axis=1)`.

ercius:
That is not right at all. Im not sure how that made it through. Im checking my test code now and will update with the actual code. I can send you a full notebook once I know it runs correctly. Sorry about that!

psavery:
@ercius Sounds good, thank you!

ercius:
I ran through my code again to verify the code below. Ill send you the full notebook as well. But it should be:
```python
[...]

PR #260, opened 2022-08-15
Add row dark algorithm to electron counting

psavery:
This adds a row dark algorithm to the electron counting. The row dark algorithm is applied to each frame pixel like the following:

```python
frame[i] *= optimizedMean / statResult;
```

Where the `optimizedMean` is obtained from the electron threshold step, and the `statResult` is either the mean or the median of the half of the row that the frame pixel is in.

This also fixes a bug where we were applying the gain before dark subtraction, when it should be the other way around (dark subtraction then gain).

This also refactors the `electronCount()` methods so that there are significantly fewer template declarations to update, and the electron counting options are now placed in structs so that, from a maintainability standpoint, it is easier to add options in the future.

Fixes: #258

### DIFF
--- a/python/image.cpp
+++ b/python/image.cpp
@@ -60,6 +60,9 @@ struct ElectronCountOptionsClassicPy
   double xRayThreshold = DBL_MAX;
   py::array_t<float> gain;
   Dimensions2D scanDimensions = { 0, 0 };
+  bool applyRowDarkSubtraction = false;
+  float optimizedMean = 0;
+  bool applyRowDarkUseMean = true;
 
   ElectronCountOptionsClassic toCpp() const
   {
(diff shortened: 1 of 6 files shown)


ercius:
I tested the row-dark algorithm using the docker image you provided. It works great!

Here is the sum of 256x256 patterns from a recent experiment counted using all three methods (image is log scale):
Figure 28

The median version took about 4x as long and it looks a little better. Is there a reason it takes so much longer?

I need to try to add the gain to this next to test it out as well. However, I dont have a good way to make a gain right now. So, lets leave it as-is for the moment.

I think this is good to merge unless there is some way to speed up the median.

psavery:
@ercius I changed the mean/median calculations to use floats instead of doubles, and that will hopefully provide a small performance increase.

I looked into other ways to improve the performance of the mean and the median, but I was unable to improve their performance. The median is a more complicated calculation than the mean, and thus it takes longer (it will typically involve some sorting and/or copying of the data, whereas the mean is just a sum and a division). Likewise in NumPy, computing the median can take a lot longer than the mean for a large array. 

I tried some alternative methods to compute the median, but they were slower. I also could not improve their performance through parallelization (for the sector reader) because the whole electron counting pipeline (including preprocessing) is already parallelized.

So I think this PR is ready!

FYI: the code to compute the mean is found [here](https://github.com/OpenChemistry/stempy/blob/e40261d8d3ab6765f405d8650a2687a1c2362438/stempy/electron.cpp#L254-L257) while the code to compute the median is found [here](https://github.com/OpenChemistry/stempy/blob/e40261d8d3ab6765f405d8650a2687a1c2362438/stempy/electron.cpp#L259-L273).

ercius:
This looks good to me. The quality of the output is very good. We should merge this in.

cjh1 (approved):
Great improvements, thank for the refactoring.

PR #263, opened 2022-08-23
Skip python 3.11 wheels

psavery:
These are not working with our current version of pybind11. Let's
skip them for now.


cjh1 (approved):
LGTM

PR #264, opened 2022-09-03
Fix com sparse parameters

ercius:
The logic in `com_sparse` was not correct when `init_center` was used and `crop_to` was not used. In this case the COM was returned as the `init_center`. This fixes that problem.

`com_sparse` will also set the COM to `nan` in cases where `crop_to` leads to no counts inside the cropped area. This matches the behavior for empty frames.

I also added a `pytest` fixture to return a simulated `SparseArray` with some variable parameters such as the size and position of the disk in diffraction space. I use this to test the new logic in `com_sparse`

@psavery How would I change `simulate_sparse_array` to have multiple frames per position? Im not sure the size of the `data` input to `SparseArray`'s `kwargs`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -116,3 +116,52 @@ def cropped_multi_frames_v2(cropped_multi_frames_data_v2):
 @pytest.fixture
 def cropped_multi_frames_v3(cropped_multi_frames_data_v3):
     return SparseArray.from_hdf5(cropped_multi_frames_data_v3, dtype=np.uint16)
+
+@pytest.fixture
+def simulate_sparse_array(scan_size, frame_size, center, how_sparse, disk_size):
+    """Make a ndarray with sparse disk.
+    scan_size: Real space scan size (pixels)
+    frame_size: detector size (pixels)
+    center: the center of the disk in diffraction space
+    how_sparse: Percent sparseness (0-1). Large number means fewer electrons per frame
+    disk_size: The radius in pixels of the diffraction disk
+    
+    """ 
+    
+#     scan_size = (100, 100)
+#     frame_size = (100,100)
+#     center = (30, 70)
+#     how_sparse = .8 # 0-1; larger is less electrons
+#     disk_size = 10 # disk radius in pixels
+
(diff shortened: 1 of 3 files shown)


tests/test_image.py
```diff
@@ -1,3 +1,5 @@
+import pytest
```
psavery, line 1: Looks like this is unused?

tests/conftest.py
```diff
@@ -116,3 +116,55 @@ def cropped_multi_frames_v2(cropped_multi_frames_data_v2):
...
+        'data': new_data,
+        'scan_shape': sparse.shape[0:2],
+        'frame_shape': sparse.shape[2:4],
+        'dtype': sparse.dtype,
+        'sparse_slicing': True,
+        'allow_full_expand': False,
+        
+    }
+    sp = SparseArray(**kwargs)
+    sp._validate()
```
psavery, line 169: I think we can skip the validate here, as it is done in `SparseArray.__init__()`.

psavery:
> I also added a `pytest` fixture to return a simulated `SparseArray` with some variable parameters such as the size and position of the disk in diffraction space. I use this to test the new logic in `com_sparse`

Sounds great!

> @psavery How would I change `simulate_sparse_array` to have multiple frames per position? Im not sure the size of the `data` input to `SparseArray`'s `kwargs`.

So, for the `data` that is passed into the `SparseArray`, if it is a 1D array, it assumed to be 1 frame per scan. If it is a 2D array, the first index is the scan position, and the second index is the frame index (for multiple frames per scan).

So if you want to have 2 frames per scan, make sure that `data.shape == (num_scan_positions, 2)`.

Alternatively, you could always use `bin_scans()` to increase the number of frames per scan.

ercius:
I updated the fixture to allow for multiple frames per position so that `data.shape == (num_scan_positions, num_frames_per_position)`. That could come in handy for testing since we dont want to pass around large H5 files.

>> Alternatively, you could always use bin_scans() to increase the number of frames per scan.

I did not know that was how `bin_scans()` worked. That is a smart way to do it!

I think this is good assuming the checks pass.

psavery (approved):
LGTM

PR #269, opened 2022-10-20
Skip incomplete frames when calculating thresholds

cjh1:


stempy/reader.h
```diff
@@ -569,7 +570,9 @@ void SectorStreamMultiPassThreadedReader::processFrames(Functor& func,
       b.header.scanDimensions = header.scanDimensions;
       b.header.imagesInBlock = 1;
       b.header.frameNumber = frameNumber;
-      b.header.imageNumbers.push_back(imageNumber);
+      b.header.imageNumbers.reserve(1);
+      b.header.imageNumbers[0] = imageNumber;
+      b.header.complete.reserve(1);
```
psavery, line 575: I think these are better to just be:

```c++
b.header.imageNumbers.push_back(imageNumber);
b.header.complete.push_back(false);
```

But let's chat about it later.

ercius:
Sorry to open this back up, but is there a way to keep track of valid frames during counting? It would be nice to have a mask written into the final HDF5 to skip such frames during post-processing the sparse data.

cjh1:
@ercius Yes, we could look at keeping track of this information. It will require a little investigation, I will raise a issue for this.

PR #272, opened 2022-11-19
Add analysis functions

ercius:
This adds 3 functions for more detailed analysis. 

`virtual_darkfield` takes in centers and radii of circular apertures to place in diffraction space. A real space virtual darkfield image is returned where each probe position is the sum of the intensity in all apertures. See below for an example

`plot_virtual_darkfield` takes in the same arguments as `virtual_darkfield` and plots circles at each aperture position to check the output is physically meaningful.
```python
center_x = (288, 317) # col, row or X, Y from matplotlib plot
center_y = (363, 213) 
radii = (15, 15)

plot_virtual_darkfield(sp, center_x, center_y, radii)
```
image

`mask_real_space` takes in a boolean array of positions matching the `scan_shape` of a `SparseArray` and outputs a summed diffraction pattern. The pattern is the sum of all positions with 1 or True in the mask. 
```python
dp_mask = mask_real_space(sp, mask)
```

image

### DIFF
--- a/python/stempy/image/__init__.py
+++ b/python/stempy/image/__init__.py
@@ -670,15 +670,12 @@ def virtual_darkfield(array, centers_x, centers_y, radii):
     :param radii: The radius of each aperture.
     :type radii: number or iterable
     
-    :param plot: If set to True then the apertures are plotted as circles using plot_virtual_darkfield
-    :rype plot: bool
-    
     :rtype: np.ndarray
     
     :example:
     >>> sp = stempy.io.load_electron_counts('file.h5')
-    >>> dp2 = stempy.image.virtual_darkfield(sp, (288, 260), (288, 160), (10, 10)) # 2 apertures
-    >>> dp1 = stempy.image.virtual_darkfield(sp, 260, 160, 10) # 1 aperture
+    >>> df2 = stempy.image.virtual_darkfield(sp, (288, 260), (288, 160), (10, 10)) # 2 apertures
+    >>> df1 = stempy.image.virtual_darkfield(sp, 260, 160, 10) # 1 aperture
     
     """
     
@@ -706,7 +703,7 @@ def plot_virtual_darkfield(image, centers_x, centers_y, radii, axes=None):
     """Plot circles on the diffraction pattern corresponding to the position and size of virtual dark field apertures.
     This has the same center and radii inputs as stempy.image.virtual_darkfield so users can check their input is physically correct.
     
-    :param image: The diffraciton pattern to plot over
+    :param image: The diffraction pattern to plot over
     :type image: np.ndarray, 2D
     
     :param centers_x: The center of each round aperture as the row locations
@@ -715,7 +712,7 @@ def plot_virtual_darkfield(image, centers_x, centers_y, radii, axes=None):
     :param centers_y: The center of each round aperture as the column locations
     :type centers_y: iterable
     
-    :param radii: The radius of each aperture. This has to be in the form (r0, )
+    :param radii: The radius of each aperture.
     :type radii: iterable
     
     :param axes: A matplotlib axes instance to use for the plotting. If None then a new plot is created.
@@ -725,7 +722,7 @@ def plot_virtual_darkfield(image, centers_x, centers_y, radii, axes=None):
     
     :example:
     >>> sp = stempy.io.load_electron_counts('file.h5')
-    >>> stempy.image.plot_virtual_darkfield(sp.sum(axis=(0,1), 260, 160, 10) # 1 aperture
+    >>> stempy.image.plot_virtual_darkfield(sp.sum(axis=(0, 1), 260, 160, 10) # 1 aperture
     """
     import matplotlib.pyplot as plt
     from matplotlib.colors import LogNorm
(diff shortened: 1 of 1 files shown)


ercius:
@psavery Is there a way to get the wheels after they are built? Then I can test this more thoroughly without having to build it myself.

cjh1:
@ercius I am happy to merge and then it will build a prerelease package that you can install.

python/stempy/image/__init__.py
```diff
@@ -653,3 +653,111 @@ def _electron_counted_metadata_to_dict(metadata):
...
+    # Change to iterable if single value
+    if isinstance(centers_x, (int, float)):
+         centers_x = (centers_x,)
+    if isinstance(centers_y, (int, float)):
+         centers_y = (centers_y,)
+    if isinstance(radii, (int, float)):
+         radii = (radii,)
+    
+    if not axes:
+        fg, ax = plt.subplots(1, 1)
```
cjh1, line 738: Where is the import for `plt`?

psavery:
The wheels that are built from the CI get uploaded inside an `artifact.zip` file that is located at the bottom of the GitHub Action run page. For the latest run, you can find them by going [here](https://github.com/OpenChemistry/stempy/actions/runs/3505288624) and scrolling to the bottom of the page (note that you must be logged into a GitHub account to see this). There will be an `artifact.zip` file that you can unzip, and then you would need to `pip install` the wheel specific for your OS and Python version.

But Chris can also create a prerelease on PyPI if you prefer that instead.

python/stempy/image/__init__.py
ercius: Just added!

ercius:
The artifact is exactly what I was looking for. No need to upload to Pypi since I can access those.

Using the artifact I found some obvious errors. Ill clean those up and let you know when its ready to review again.

ercius:
@psavery I think this is ready to review before merging.

python/stempy/image/__init__.py
```diff
@@ -653,3 +653,116 @@ def _electron_counted_metadata_to_dict(metadata):
...
+
+    rs_image = np.zeros((array.shape[0] * array.shape[1],), dtype=array.dtype)
+    for ii, events in enumerate(array.data):
+        for ev in events:
+            ev_rows = ev // array.frame_shape[0]
+            ev_cols = ev % array.frame_shape[1]
+            for cc_0, cc_1, rr in zip(centers_x, centers_y, radii):
+                dist = np.sqrt((ev_rows - cc_1)**2 + (ev_cols - cc_0)**2)
+                rs_image[ii] += len(np.where(dist < rr)[0])
+    rs_image = rs_image.reshape(array.shape[0:2])
```
psavery, line 701:
```suggestion
    rs_image = rs_image.reshape(array.scan_shape)
```

python/stempy/image/__init__.py
```diff
@@ -653,3 +653,116 @@ def _electron_counted_metadata_to_dict(metadata):
         ret[name] = getattr(metadata, name)
 
     return ret
+
+def virtual_darkfield(array, centers_x, centers_y, radii, plot=False):
```
psavery, line 657: Looks like the `plot` argument isn't actually being used right now?

python/stempy/image/__init__.py
```diff
@@ -653,3 +653,116 @@ def _electron_counted_metadata_to_dict(metadata):
...
+    axes.imshow(image, cmap='magma', norm=LogNorm())
+    
+    # Place a circle at each apertue location
+    for cc_0, cc_1, rr in zip(centers_x, centers_y, radii):
+        C = Circle((cc_0, cc_1), rr, fc='none', ec='c')
+        axes.add_patch(C)
+    
+    return axes
+
+def mask_real_space(array, mask):
```
psavery, line 750: Nice! This can potentially also be a method on the `SparseArray` itself.

python/stempy/image/__init__.py
```diff
@@ -653,3 +653,116 @@ def _electron_counted_metadata_to_dict(metadata):
...
+    
+    """
+    
+    # Change to iterable if single value
+    if isinstance(centers_x, (int, float)):
+         centers_x = (centers_x,)
+    if isinstance(centers_y, (int, float)):
+         centers_y = (centers_y,)
+    if isinstance(radii, (int, float)):
+         radii = (radii,)
```
psavery, line 691: If there's any chance the single values could be a numpy type (like `np.int32`), then this check might not work. But we could also just wait and see if we run into that issue.

psavery (approved):
I left a few comments. Overall, looks good! Very nice looking images too!

python/stempy/image/__init__.py
ercius: I removed it. It seemed convenient at the time, but hard to implement in practice

python/stempy/image/__init__.py
ercius: someone might use an `numpy.ndarray`. Im not sure the best way to test for all of these things. Lets see how it goes as you suggest.

python/stempy/image/__init__.py
ercius: I agree. This sort of allows for "fancy indexing". Maybe the same algorithm can be used for that functionality.

psavery (approved):
Looks good to me!

Issue #267, opened 2022-09-27
question: SparseArray fancy indexing?


ercius:
@psavery Is there a way to do fancy indexing in the `SparseArray` class? I want to create a mask of an arbitrary shape and get the summed diffraction pattern from that mask.

ndarray fancy indexing example:
```python
ee = np.eye(3)
ee[(0, 1, 2), (0, 1, 2)]
```
array([1., 1., 1.])

This throws a value error for a SparseArray though:
```python
sa = SparseArray.from_hdf5('filename.h5')
sa2 = sa[(0, 1, 2), (0, 1, 2), :, :]
```

```
ValueError                                Traceback (most recent call last)
<ipython-input-27-e4191eacb753> in <module>
      4 xx,yy = np.nonzero(mask)
      5 
----> 6 sa2 = sa[xx[0:2],yy[0:2],:,:]

/opt/tljh/user/envs/stempy/lib/python3.8/site-packages/stempy/io/sparse_array.py in __getitem__(self, key)
    658         for i, item in enumerate(key):
    659             if not isinstance(item, slice):
--> 660                 if item >= self.shape[i] or item < -self.shape[i]:
    661                     raise IndexError(f'index {item} is out of bounds for '
    662                                      f'axis {i} with size {self.shape[i]}')

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

A `for` loop works but is there a way to do this closer to the fancy indexing version? I see how returning a SparseArray seems awkward since it does not really have a proper `scan_shape`.

The use case is in returning a summed frame from the indicated positions:

```python
dp = np.zeros(sa.frame_shape, dtype=sa.dtype)
xx, yy = np.indices((10, 10))
for ii in xx.ravel():
    for jj in yy.ravel():
        dp += sa[ii, jj, :, :]
```

psavery:
Hi @ercius. I think advanced indexing (with integer arrays and boolean arrays) can certainly be possible with the `SparseArray` class, but it would need to be implemented.

[...]

PR #276, opened 2023-03-07
Sparse array advanced indexing

psavery:
This adds advanced indexing to sparse arrays.

Most advanced indexing is allowed, with the following exceptions:

1. You can't perform advanced indexing on both the scan positions and frame positions in the same operation simultaneously. So this isn't allowed: `array[[0, 1, 2], [0, 1, 2], [0, 1, 2], [0, 1, 2]]`. This is not allowed because we perform indexing on the scan positions and frame positions separately, and performing advanced indexing on both simultaneously is complicated (and may not be needed either).
2. 2D advanced indexing is only allowed for masks (i. e., boolean arrays with the same shape as the scan/frame). This is because other types of 2D advanced indexing can add extra dimensions, which we don't want to do here.

Additionally, the option to pass `axis='scan'` or `axis='frame'` was added to the arithmetic operations. This allows the user to perform advanced indexing followed by an arithmetic operation without having to figure out how the axes have changed. For instance: `array[[0, 1, 2], [0, 1, 2]].sum(axis='scan')`.

Fixes: #267

### DIFF
--- a/tests/test_sparse_array.py
+++ b/tests/test_sparse_array.py
@@ -632,6 +632,44 @@ def test_advanced_indexing(sparse_array_small, full_array_small):
     assert np.allclose(array[scan_mask].mean(axis='scan'),
                        full[scan_mask].mean(axis=(0,)))
 
+    # Now do some basic tests with multiple frames per scan
+    data = np.empty((2, 2), dtype=object)
+    data[0][0] = np.array([0])
+    data[0][1] = np.array([0, 2])
+    data[1][0] = np.array([0])
+    data[1][1] = np.array([0, 1])
+    kwargs = {
+        'data': data,
+        'scan_shape': (2, 1),
+        'frame_shape': (2, 2),
+        'sparse_slicing': False,
+        'allow_full_expand': True,
+    }
+    # The multiple frames per scan array
+    m_array = SparseArray(**kwargs)
+
+    # These are our expected expansions of the positions
+    position_zero = np.array([[2, 0], [1, 0]])
+    position_one = np.array([[2, 1], [0, 0]])
+
+    # Verify our assumption is true
+    assert np.array_equal(m_array[0, 0], position_zero)
+    assert np.array_equal(m_array[1, 0], position_one)
+
+    # Now test some simple fancy indexing
+    assert np.array_equal(m_array[[0], [0]][0], position_zero)
+    assert np.array_equal(m_array[[0, 1], [0]][0], position_zero)
+    assert np.array_equal(m_array[[0, 1], [0]][1], position_one)
+    assert np.array_equal(m_array[[1, 0], [0]][0], position_one)
+    assert np.array_equal(m_array[[1, 0], [0]][1], position_zero)
+    assert np.array_equal(m_array[0, 0, [0, 1], [0, 0]], [2, 1])
+    assert np.array_equal(m_array[0, 0, [[True, False], [True, False]]],
+                          [2, 1])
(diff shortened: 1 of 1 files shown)


tests/test_sparse_array.py
```diff
@@ -514,6 +514,105 @@ def test_data_conversion(cropped_multi_frames_v1, cropped_multi_frames_v2,
...
+        [[3, 2, 1], [0, 1, 2]],
+        [[-1, 4, -3], [-3, 8, 7]],
+        [[7], 4],
+        [3, [8]],
+        [[-1, 4, -3], slice(2, 18, 2)],
+        [3, 4, [3, 2, 1], [9, 8, 7]],
+        [slice(None), slice(None), [5, 45, 32], [-3, -7, 23]],
+        [slice(None), slice(2, 18, 2), slice(0, 100, 3), [-5, 1, 2]],
+    ]
+
```
psavery, line 536: @ercius You can see some of the advanced integer indexing test cases we are running above.

tests/test_sparse_array.py
```diff
@@ -514,6 +514,105 @@ def test_data_conversion(cropped_multi_frames_v1, cropped_multi_frames_v2,
...
+    # Now test boolean indexing.
+    # We'll make some random masks for this.
+    rng = np.random.default_rng(0)
+
+    num_tries = 3
+    for i in range(num_tries):
+        scan_mask = rng.choice([True, False], array.scan_shape)
+        frame_mask = rng.choice([True, False], array.frame_shape)
+
```
psavery, line 574: @ercius For testing the masking, we are creating random masks for the scans and the frames and applying them.

tests/test_sparse_array.py
ercius: these all look like possible user cases to me. I think this will make our code much more compatible with other applications like py4DSTEM.

tests/test_sparse_array.py
ercius: The most common case of indexing in the frame dimension would be simple indexing like `sparse_array[:,:,100:200,200:300]` where you want to crop out part of the frames. So, testing and enabling that is the most important.

tests/test_sparse_array.py
psavery: Sounds great!

tests/test_sparse_array.py
psavery: I believe this should already be supported (since this is regular indexing). I will add a test that specifically does this so that we can verify!

psavery: Indeed it does appear to work! I just added a specific test for it so we can ensure it stays working.

tests/test_sparse_array.py
```diff
@@ -97,6 +97,26 @@ def test_sparse_slicing(sparse_array_small, full_array_small):
...
+    # First test the sparse slicing version
+    sliced = array[:, :, 100:200, 200:300]
+    sliced.sparse_slicing = False
+
+    assert np.array_equal(sliced[:], full[:, :, 100:200, 200:300])
+
+    # Now test the dense slicing version
+    array.sparse_slicing = False
+    assert np.array_equal(array[:, :, 100:200, 200:300],
+                          full[:, :, 100:200, 200:300])
```
psavery, line 117: @ercius Here's that test I just added.

cjh1 (commented):
LGTM

cjh1:
@ercius These changes are now in 3.2.3

Issue #277, opened 2023-03-10
Summing in frame space error


ercius:
Summing in frame space throws up an error if `allow_full_expand == False`. I dont think this is necessary since summing in frame space just has to do with counting the total number of electrons in that frame.

So, this throws an error:
```python
import stempy.io as stio
sp = stio.load_electron_counts('file.h5') 
sp2[:,:,0:100,0:100].sum(axis=(2,3))
```
FullExpansionDenied: Full expansion is not allowed

Can the sum be changed to simply `len()` for each frame location or am I missing something?

Something like:
```python
frame_sum = np.zeros((sp.scan_shape[0] * sp.scan_shape[1]))
for ii, pos in enumerate(sp2.data):
    for frame in pos:
        frame_sum [ii] += len(frame)
````

**Comment by psavery** [comment]
Sure @ercius! I can look into that.

The behavior of the `SparseArray` right now for the math operations is basically:

1. Check if there is a special implementation for the selected axes
2. If there's no special implementation, then perform a full expansion and perform the operation on the full array

Right now, there are only special implementations for performing math operations along the scan axes. But I can add implementations for the frame axes too!

For now, I can at least add an implementation for summing along the frame axes. And we can add implementations for the other math operations too as needed.

**Comment by psavery** [comment]
@ercius I could see us adding something similar for `min()` and `max()` by changing the line:
```python
        frame_sum[ii] += len(frame)
```

To be
```python
    frame_min[ii] = min(len(frame), frame_min[ii])
```

(and the same kind of thing for `max()`).

Should we add implementations for `min` and `max` also?

ercius:
Im not sure this is the same as the `min()` and `max()` for a regular `ndarray`. I think that calculates the number of electrons on of the frame with the fewest electrons. That is potentially useful but not the same as `min(ndarray)`. I think its more like `ndarray.sum().min()`.

psavery:
@ercius You were right! It wasn't that simple. But I took a quick look and figured out how to implement them (and added them to #278).

I'm not sure how useful the `min()` would be, since it is almost always all zeros. For one frame per scan, the `max()` is almost always all ones. But maybe the `max()` will be more useful for multiple frames per scan (where it might be 2 or more).

ercius:
This is still useful in that it make the `SparseArray` class more compatible with `ndarray`s. Thanks for implementing that!

PR #278, opened 2023-03-10
Add implementation for summing along frame axes

psavery:
Several tests were added to verify that this is working correctly. These include tests for multiple frames per scan position, slicing followed by a frame axes sum, and advanced indexing followed by a frame axes sum.

Since the mean is implemented using the sum, this addition also adds support for performing a mean along the frame axes as well.

Fixes: #277


ercius:
I just looked and that looks good to me! You are right that it will usually be 0 and 1 for min and max; howver, this allows other code to call the `min()` and `max()` functions without erroring. So, it makes `SparseArray` a more complete clone of `ndarray`. Great work!

cjh1 (approved):
LGTM

Issue #279, opened 2023-03-21
Implement reader that allows reading specific frames


cjh1:
The idea would be to have a version on the `SectorStreamMultiPassThreadedReader` that would allow the user to read just a set of frames. This could be implemented by first called [readHeaders](https://github.com/OpenChemistry/stempy/blob/master/stempy/reader.h#L530) to create the scan map. The API would then allow the user to specify a range of frames that they want to read at a give scan position. To implement this method we will need to refactor [processFrame](https://github.com/OpenChemistry/stempy/blob/master/stempy/reader.h#L549) so that we have a way of read just a set of frames from the map. The idea is the Python API would look something like this:

```python
import stempy.io as stio
reader = stio.reader(iFiles,stio.FileVersion.VERSION5, backend="multi-pass-partial") 
frames = reader.read_frames(scan_position, frame_slice)
```

Note: The first time `read_frames` is calling it would call `readHeader`, subsequent call can just reused the scan map. The reading of the frames can be done in a threadpool or just on a single thread, the idea is that it should be a subset.

ercius:
Does the reader object contain any information about the scan? E.g. does `reader.scan_shape` provide the number of frames?

cjh1:
> Does the reader object contain any information about the scan? E.g. does `reader.scan_shape` provide the number of frames?

No, that information is in the header, so you have to read a block.

cjh1:
@ercius When the user specifies which frames they are interesting in would they uses the frameNumber as found in the header or the index ( as in 0, would be the first frame at the position, regardless of the frameNumer ).

psavery:
> @ercius When the user specifies which frames they are interesting in would they uses the frameNumber as found in the header or the index ( as in 0, would be the first frame at the position, regardless of the frameNumber ).

@ercius FYI: the example I'm working with from `2022.01.13` has frame numbers like `1490148420` (each frame has its own unique number, and they all appear to be 10 digits).

ercius:
Those ten digit frame numbers will not be useful to a user. I think those
are global frame numbers from the camera only useful for stitching sectors
into frames.

psavery:
@ercius Just double-checking something: will the returned frames be electron counted, or raw frames?

ercius:
Just raw frames.

PR #280, opened 2023-03-29
Add method for reading frames at a scan position

psavery:
This refactors the `SectorStreamMultiPassThreadedReader` class a little bit and adds methods for reading specific frames from the data.

Reading the frames requires first reading all of the headers of the data (which can take some time, since we have to parse the data files). But once the headers have been read once, they are cached so that they do not need to be read again.

Here is some example code:

```python
import glob

import stempy.io as stio

backend = 'multi-pass'
files = glob.glob('*.data')

reader = stio.reader(files, stio.FileVersion.VERSION5, backend=backend)

blocks = reader.read_frames((5, 30), [1, 2, 3])

for block in blocks:
    array = block.data  # this is the numpy array for the block
```

The `frames_slice` argument is used to slice directly into a `np.arange(num_frames)` array, and the result is used to determine which frames, and in what order, get returned.

A couple of additional methods were added for convenience:

1. `reader.scan_shape` will return the scan shape (if a header hasn't been read yet, it will automatically read one header to determine this)
2. `reader.num_frames_per_scan` will return the number of frames per scan. This must read all of the headers, and can be time consuming if that hasn't been done yet (but reading all of the headers will only happen once per reader object).

Fixes: #279

### DIFF
--- a/python/io.cpp
+++ b/python/io.cpp
@@ -126,5 +126,10 @@ PYBIND11_MODULE(_io, m)
     .def(py::init<const std::string&, int>(), py::arg("path"),
          py::arg("threads") = 0)
     .def(py::init<const std::vector<std::string>&, int>(), py::arg("files"),
-         py::arg("threads") = 0);
+         py::arg("threads") = 0)
+    .def("_load_frames", &SectorStreamMultiPassThreadedReader::loadFrames)
+    .def("_num_frames_per_scan",
+         &SectorStreamMultiPassThreadedReader::numFramesPerScan)
+    .def("_scan_dimensions",
+         &SectorStreamMultiPassThreadedReader::scanDimensions);
 }
--- a/python/stempy/io/__init__.py
+++ b/python/stempy/io/__init__.py
@@ -88,8 +88,8 @@ class SectorThreadedMultiPassReader(ReaderMixin, _threaded_multi_pass_reader):
     def read_frames(self, scan_position, frames_slice=None):
         """Read frames from the specified scan position and return them
 
-        The scan_position must be a tuple of a valid position in the
-        scan_shape.
+        The scan_position is either a tuple of a valid position in the
+        scan_shape, or an integer that is a flattened index of the position.
 
         The frames_slice object will be used as an index in numpy to
         determine which frames need to be read. If None, all frames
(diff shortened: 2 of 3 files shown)


ercius:
When are the headers read? Is it when the reader is created or when you call `read_frames`?

I dont quite understand the docstring. Is the `frame_slice` meant to be there to accommodate cases where `reader.num_frames_per_scan` is > 1?

psavery:
> When are the headers read? Is it when the reader is created or when you call `read_frames`?

The headers are lazily read when they first need to be read. If you call `reader.num_frames_per_scan`, it will read them (because it must in order to determine this). If you call `reader.read_frames()` and the headers haven't been read yet, they will be read then.

> I dont quite understand the docstring. Is the `frame_slice` meant to be there to accommodate cases where `reader.num_frames_per_scan` is > 1?

Yes. It is following that design Chris gave in #279, where you select the frames you want to extract. For instance, `0`, or `[1, 2, 3]`, or `[2, -1]`. It can be anything that can be used to slice into a numpy array (and it is done like `np.arange(num_frames_per_scan)[frame_slice]`).

psavery:
> Ok great. I have two questions/comments though:
> 
> 1. If I use `reader.read_frames((0, 5000))` there is no error even though the scan shape is (64, 256). **This is desirable**, because in some cases I want to think of a "scan" in terms of the raveled (flattened) shape instead of a 2D scan shape. But its confusing the way it works currently. Can you make it so that `arg0` as a tuple assumes you want a specific scan position and `arg0` as a single number assumes you want to index into the scan using the raveled frame number (i.e. same as `(0, frame_num)`)?

This sounds great to me!

> 2. Is there a way to get a range of frames? Or should I just make a for loop outside?

Yes. You can pass a slice object as the second argument, like `reader.read_frames((5, 30), slice(3, 7))` (that will get all frames from 3 to 7).

Let me know if you have any ideas on how to simplify this API. It would have been nice to be able to do things like `frames[3:7]`, but that slicing can only be done on an object (we can't do `reader.read_frames((5, 30), 3:7)`). Although you could do it like this instead:
```python
frames = np.arange(num_frames)[3:7]
reader.read_frames((5, 30), frames)
```

Also, the function is returning a list of blocks. This was in case you wanted to extract any information from, for instance, the `block.header`. But if the block information doesn't matter at all, we could just create a concatenated numpy array and return it instead.

psavery:
@ercius I just pushed a commit, you can now specify the scan position as a flattened index instead!

ercius:
That works great. I think we are good to go with this.

psavery:
Awesome! @cjh1 This is ready for review, then!

cjh1 (approved):
LGTM, some nice improvements, plus new functionality!

psavery:
> LGTM, some nice improvements, plus new functionality!

Thanks Chris!

cjh1:
@ercius I plan on cutting 3.3.3 to include this new feature, if you have no objections.

ercius:
Please do!

cjh1:
Done

cjh1:
@ercius I have updated the docker images, let me know if you run into any problems.

PR #282, opened 2023-04-28
avoid type conversion to float on h5 import

swelborn:
This is a small PR that avoids type conversion behavior to float on integer division.

@ercius and I (mostly @ercius) figured out that this line:

```python
data = data.reshape((np.prod(scan_shape), frames_per_scan))
```

Fails with:

```python
TypeError: 'numpy.float64' object cannot be interpreted as an integer
```

When the type of the `scan_shape` list is an unsigned int, `np.prod(scan_shape)`, and integer division results in a float64:

```python 
frames_per_scan = len(data) // np.prod(scan_shape) # (python int) // np.uint 
print(frames_per_scan.dtype)
# float64
```
This change ensures that this conversion doesn't happen.

### DIFF
--- a/python/stempy/io/sparse_array.py
+++ b/python/stempy/io/sparse_array.py
@@ -1,8 +1,8 @@
-from collections.abc import Sequence
 import copy
-from functools import wraps
 import inspect
 import sys
+from collections.abc import Sequence
+from functools import wraps
 
 import numpy as np
 
@@ -202,8 +202,10 @@ class SparseArray:
         scan_shape = scan_shape[::-1]
 
         if version >= 3:
+            # Convert to int to avoid integer division that results in 
+            # a float
+            frames_per_scan = len(data) // int(np.prod(scan_shape))
             # Need to reshape the data, as it was flattened before saving
-            frames_per_scan = len(data) // np.prod(scan_shape)
             data = data.reshape((np.prod(scan_shape), frames_per_scan))
 
         # We may need to convert the version of the data
--- a/python/stempy/io/sparse_array.py
+++ b/python/stempy/io/sparse_array.py
@@ -1,8 +1,8 @@
+from collections.abc import Sequence
 import copy
+from functools import wraps
 import inspect
 import sys
-from collections.abc import Sequence
-from functools import wraps
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


python/stempy/io/sparse_array.py
```diff
@@ -1,8 +1,8 @@
-from collections.abc import Sequence
 import copy
-from functools import wraps
 import inspect
 import sys
+from collections.abc import Sequence
+from functools import wraps
```
cjh1, line 6: Any reason these moved?

psavery:
> When the type of the `scan_shape` list is an unsigned int, `np.prod(scan_shape)`, and integer division results in a float64:

Nice find! I was able to reproduce this issue with the following in the terminal:
```python
>>> np.prod(np.array((3, 5), dtype=np.uint64)).dtype
dtype('uint64')
>>> (30 // np.prod(np.array((3, 5), dtype=np.uint64))).dtype
dtype('float64')
```
It appears that if there is a numpy operation between signed and unsigned integers, it promotes the type to a float.
```python
>>> (np.int64(30) + np.uint64(1)).dtype
dtype('float64')
```

python/stempy/io/sparse_array.py
swelborn: That was my format on save, fixed.

python/stempy/io/sparse_array.py
```diff
@@ -202,8 +202,10 @@ def from_hdf5(cls, filepath, **init_kwargs):
         scan_shape = scan_shape[::-1]
 
         if version >= 3:
+            # Convert to int to avoid integer division that results in 
+            # a float
+            frames_per_scan = len(data) // int(np.prod(scan_shape))
```
cjh1, line 207: `np.prod(scan_shape, dtype=int)` would also work I believe.

cjh1 (approved):
LGTM, just waiting on CI

python/stempy/io/sparse_array.py
swelborn: changed.

cjh1:
@ercius @swelborn These changes are now in 3.3.4

PR #284, opened 2023-05-09
Add get scan path

ercius:
This adds a function to return the path to a counted data set using the `scan_id`, `scan_num`, and/or `threshold` to `io`.


cjh1:
@ercius I can see the utility of this function, however, I wonder if it might be better off in another location that is more site specific. I worry about encoding distiller conventions in stempy. Maybe included inline in the notebooks? Another alternative would be to put it in a `contrib` directory, indicating that its not part of the stempy public interface. I am open to other options too.

ercius:
>I wonder if it might be better off in another location that is more site specific. I worry about encoding distiller conventions in stempy.

That makes sense. Im happy to move it anywhere that makes sense.

> Maybe included inline in the notebooks? 

I was trying to avoid this because it adds unnecessary code and looks really cluttered. Maybe that is just something to live with? It is annoying to copy this between notebooks and makes it more prone to getting out of sync across notebooks.

>Another alternative would be to put it in a contrib directory, indicating that its not part of the stempy public interface.

I think that would work too. It would be a simple import in the notebooks that need it. This seems like the best trade-off to me.

ercius:
After our recent discussion I think `contrib` is a great alternative. Ill update the PR.

cjh1 (approved):

PR #285, opened 2023-05-10
Use BigMPI to send events

cjh1:
With the new detector and multiple frames per scan position we are starting to get more events than can be sent in a regular MPI message. This adds BigMPI to increase the size of the messages we can sent to get around this restriction.


psavery (approved):
LGTM

PR #289, opened 2023-06-02
New PR for GHA

swelborn:
After discussion with @cjh1, this is a cleaner PR for GHA.

I tried to strike a middle ground:
1. Added scripts to build locally.
2. "latest" and the latest commit tag are aliased.
3. Remove conda in favor of python docker images and pip.

"py3x" is still in the tag name, but it is also "latest" without any python version name. Had to do this to ensure base build maps to stempy build. Let me know what you think about these changes.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,137 @@
...
+          echo "IPYKERNEL_TAG=${IPYKERNEL_TAG}" >> $GITHUB_ENV
+
+          BASE_TAG=${{ env.DOCKERHUB_ORG }}/stempy${MPI_TAG}:py${{ env.PYTHON_VERSION_NODOT }}-base${DEV_TAG}
+          echo "BASE_TAG=${BASE_TAG}" >> $GITHUB_ENV
+
+          TAG=${{ env.DOCKERHUB_ORG }}/stempy${MPI_TAG}${IPYKERNEL_TAG}:py${{ env.PYTHON_VERSION_NODOT }}-${COMMIT}${DEV_TAG}
+          echo "TAG=${TAG}" >> $GITHUB_ENV
+
+          LATEST_TAG=${{ env.DOCKERHUB_ORG }}/stempy${MPI_TAG}${IPYKERNEL_TAG}:latest${DEV_TAG}
+          echo "LATEST_TAG=${LATEST_TAG}" >> $GITHUB_ENV
```
psavery, line 103: If you have more than a few lines in a `run` command, you may want to consider placing the code in a script instead, and calling it. But you do not have to.

.github/actions/docker_setup/action.yml
psavery, line 1: This looks like it is mostly just two actions combined into one, right? What are the benefits for creating a new action for this?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,137 @@
...
+          DOCKERHUB_ORG: ${{ vars.DOCKERHUB_ORG }}
+
+      - name: Determine if Dockerfile.base changed
+        id: changed-dockerfile-base
+        uses: tj-actions/changed-files@v35
+        with:
+          files: |
+            ./docker/Dockerfile.base
+            ./docker/apt-packages-common.txt
+            ./docker/apt-packages-dev.txt
```
psavery, line 48: Does the `Dockerfile.base` take a long time to build? If not, it could have been simpler to just build it every time, rather than caching it. We may actually want to rebuild it occasionally just because upstream dependencies change over time which include things like security updates.

But we can leave this!

psavery (approved):
I left a few comments, but overall, I think it looks **much** simpler than before, and it looks great!

.github/workflows/docker.yml
swelborn: Dockerfile.base does take a long time to build (~40 minutes). I understand that it is CI, but say you are developing on a fork and want to make a quick change and test at NERSC. 

All you have to do is edit this:
```yml
on:
  push:
    branches:
      - 'master'
  pull_request:
    branches:
      - 'master'
```

And commit your change, and CI will take care of building/pushing to your own Dockerhub repo (so long as you have set up these vars and secrets):
```
DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
DOCKERHUB_ORG: ${{ vars.DOCKERHUB_ORG }}
```

I can set up a cron job to build/push base, I don't think that would be too much work.

psavery: Thanks for the info! It's probably okay for now unless @cjh1 thinks we should add a cron job.

.github/actions/docker_setup/action.yml
swelborn: In previous iterations I was using this composite action several times throughout the workflow. I can remove it. I do think it makes it a little easier to read, and you could reuse it later. Let me know your opinion!

swelborn:
@cjh1 I tested `samwelborn:stempy-mpi:latest` and it works. All of the distiller notebooks also work with `samwelborn:stempy-ipykernel:latest` Should be good to go, if you are ok with it in its current state.

PR #290, opened 2023-07-13
contrib: add typing, include versions in get_scan_path

swelborn:
improves on @ercius commit to enable getting scan path for new naming convention.

adds typing, and improves flow


ercius (approved):
LGTM

Issue #292, opened 2023-08-04
BigMPI causes Cereal error


swelborn:
This error occurs with what seems to be random times. 

```bash
RuntimeError: Failed to read 836 bytes from input stream! Read 132
```

I ran the following command on the distiller jobs folder:

```bash
grep -rl --include="*.out" 'RuntimeError: Failed to read' /global/cfs/cdirs/ncemhub/distiller/jobs/ > ~/RuntimeError_jobs.txt
```

This is the output:
<details>

<summary>RuntimeError_jobs.txt</summary>

/global/cfs/cdirs/ncemhub/distiller/jobs/6187/slurm-13083213.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6388/slurm-13200700.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6173/slurm-13056663.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6133/slurm-11488771.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6327/slurm-13170411.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6254/slurm-13125660.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6448/slurm-13222002.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6455/slurm-13223008.out
/global/cfs/cdirs/ncemhub/distiller/jobs/5387/slurm-2674179.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6375/slurm-13194823.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6240/slurm-13122082.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6200/slurm-13094426.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6196/slurm-13085878.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6500/slurm-13238649.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6496/slurm-13238069.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6182/slurm-13082172.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6142/slurm-11525534.out
/global/cfs/cdirs/ncemhub/distiller/jobs/5388/slurm-2674240.out
/global/cfs/cdirs/ncemhub/distiller/jobs/6241/slurm-13122090.out
[...]

PR #293, opened 2023-08-16
Big MPI fix

swelborn:
See #292 for details on this fix. 

I think the sends/receives were mixing up the message size message and the data messages, because of MPI_ANY_SOURCE.

I tested this and it works.

### DIFF
--- a/stempy/electron_mpi.cpp
+++ b/stempy/electron_mpi.cpp
@@ -34,21 +34,22 @@ void serializeBlocks(std::vector<Block>& blocks, std::ostream* stream)
   archive(blocks);
 }
 
-void receivePartialMap(Events& events, std::vector<char>& recvBuffer)
+void receivePartialMap(Events& events, std::vector<char>& recvBuffer,
+                       int sourceRank)
 {
   MPI_Status status;
 
-  // First get the message size
+  // First get msg size using the source rank as the tag
   size_t msgSize;
-  MPI_Recv(&msgSize, 1, MPI_UINT64_T, MPI_ANY_SOURCE, 0, MPI_COMM_WORLD,
+  MPI_Recv(&msgSize, 1, MPI_UINT64_T, sourceRank, sourceRank, MPI_COMM_WORLD,
            MPI_STATUS_IGNORE);
 
   // Resize our buffer
   recvBuffer.resize(msgSize);
 
   // Now receive the data using BigMPI
-  MPIX_Recv_x(recvBuffer.data(), recvBuffer.size(), MPI_BYTE, MPI_ANY_SOURCE, 0,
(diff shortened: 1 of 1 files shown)


psavery (commented):
@swelborn Great job tracking this bug down! Race conditions like this can sometimes be very difficult to find!

The only change that I would suggest is: can you try setting all of the `tag` arguments (both for sends and receives) back to `0`? I think specifying the `source` rank in the `MPI_RECV` calls should be enough. Let me know if that works when you try it out.

These changes might result in some reduced performance since rank `0` must now receive all partial maps in order of rank, instead of the order in which they are sent. So rank `0` might end up having more idle time. But this might also be insignificant (@cjh1 would know better than me). It's more important for it to run correctly. Another possible solution might be to perform an `MPI_Probe` before the first `MPI_Recv` in order to obtain the rank of the incoming source, and then specify that source rank in the following `MPI_Recv` and `MPI_Recvx` calls. But if the performance difference is negligible, then it doesn't matter. We can merge this PR anyways and try that out later.

After you try setting all of the tag arguments back to `0`, this will be ready to merge!

swelborn:
@psavery Thanks for your review. Yes, I can make the change for the tags.

I totally understand your point about `rank 0` waiting around a bit around more, but I don't think this is a significant issue. We could get more fancy with it, but right now there are only 4 ranks in production... if we were to (for some reason) expand to more ranks, it may be worth it to try these other things out.

I will replace the tags sometime next week and test.

swelborn:
I tested this, and it shows normal output:

```
****Statistics for calculating electron thresholds****
number of samples:1200
min sample:0
max sample:1685
mean:20.6932
variance:5.81956
std dev:2.41238
number of bins:1089
x-ray threshold n sigma:175
background threshold n sigma:4.5
optimized mean:20.6288
optimized std dev:1.78833
background threshold:28.6763
xray threshold:442.859
electron_count_cori.py
Using files in /pscratch/sd/s/swelborn/ncem/2023.05.27
scan name = data_scan0000000440_*.data
start counting #440
electron_count_cori.py
Using files in /pscratch/sd/s/swelborn/ncem/2023.05.27
scan name = data_scan0000000440_*.data
start counting #440
electron_count_cori.py
Using files in /pscratch/sd/s/swelborn/ncem/2023.05.27
scan name = data_scan0000000440_*.data
start counting #440
electron_count_cori.py
Using files in /pscratch/sd/s/swelborn/ncem/2023.05.27
scan name = data_scan0000000440_*.data
start counting #440
total time = 10.152918815612793
/pscratch/sd/s/swelborn/ncem/2023.05.27/data_scan440_th4.5_electrons.h5
```

psavery (approved):
LGTM

Issue #294, opened 2023-08-16
Python 3.11 installs stempy 0.0.1


ercius:
pip will only install stempy version 0.0.1 if you are using python 3.11. Im not sure if 3.11 is supported. The install files need to be updated to allow install on 3.11.

PR #295, opened 2023-08-17
Build wheels for Python 3.11 also

psavery:
I believe Python 3.11 wheels were being skipped because, at the time, several dependencies did not have Python 3.11 support.

But our dependencies should have Python 3.11 support now. So let's add back in Python 3.11.

Fixes: #294

### DIFF
--- a/.github/scripts/docker/Dockerfile
+++ b/.github/scripts/docker/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=quay.io/pypa/manylinux2010_x86_64
+ARG BASE_IMAGE=quay.io/pypa/manylinux2014_x86_64
 FROM ${BASE_IMAGE}
 
 RUN yum install -y \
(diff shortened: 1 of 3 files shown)


psavery:
@cjh1 I updated the dockerhub wheelbuilder to use manylinux2014 instead of manylinux2010. This is because [manylinux2010 is no longer supported as of about a year ago](https://github.com/pypa/manylinux#docker-images). manylinux2010 also does not contain python3.11, but manylinux 2014 does.

psavery:
After adding Python 3.11, upgrading our docker image to manylinux2014, and upgrading pybind11 to the latest version, building the Python 3.11 wheels now appears to work!

@ercius feel free to download a Python 3.11 wheel from [here](https://github.com/OpenChemistry/stempy/actions/runs/5889070122) to give it a try. I hope upgrading pybind11 didn't introduce any issues.

Once this PR is approved, we can merge it, and then push a tag, and then Python 3.11 wheels should be available on PyPI.

swelborn:
looks good.

@ercius should still check if all functionality is still there after upgrading pybind11.

cjh1 (approved):
LGTM

PR #298, opened 2023-10-03
Try cross-compliing to get wheels for Apple silicon

cjh1:


cjh1:
@psavery Looks like this successfully build the Apple Silicon wheels.

psavery (approved):
Wow, that's great news! Looks like they built! Let me know if you want me to try one of them out before merging.

cjh1:
@psavery I will cut a point release, then we can test it out.

PR #299, opened 2023-10-26
Bump HDF5 build version on the Mac

cjh1:


psavery:
Were these changes specifically for arm64, x86_64, or both?

cjh1:
For both, although I don't have access to a x86_64 mac to test with.

psavery (approved):
I gave a wheel a quick try on both x86_64 mac and arm64 mac, and it worked on both! Looks good to me!

cjh1:
Thanks @psavery for the review and testing!

PR #301, opened 2023-10-26
Remove cmake from requirements.txt

cjh1:
cmake is not needed at runtime ...


swelborn (approved):
that looks good, as long as things build!

PR #304, opened 2023-11-06
Remove flyback

ercius:
Update `from_hdf5` to allow the user to crop the flyback column during load. For large data sets this can reduce the time needed in processing later significantly. The keyword is set to `True` to be compatible with older code. 

Need to fully test once the CI builds are available. Dont merge yet. Will update as needed.

### DIFF
--- a/python/stempy/io/sparse_array.py
+++ b/python/stempy/io/sparse_array.py
@@ -197,24 +197,32 @@ class SparseArray:
                 data = frames[()] # load the full data set
                 scan_positions = scan_positions_group[()]
             else:
+                print(scan_shape)
                 # Generate the original scan indices from the scan_shape
                 orig_indices = np.ravel_multi_index([ii.ravel() for ii in np.indices(scan_shape)],scan_shape)
+                print(orig_indices.shape)
+                print(orig_indices)
                 # Remove the indices of the last column 
-                crop_indices = np.delete(orig_indices, orig_indices[::scan_shape[1]])
+                crop_indices = np.delete(orig_indices, orig_indices[scan_shape[0]-1::scan_shape[0]])
+                print(crop_indices.shape)
+                print(crop_indices)
                 # Load only the data needed
                 data = frames[crop_indices]
+                print(data.shape)
                 # Reduce the column shape by 1
-                scan_shape[1] = scan_shape[1] - 1
+                scan_shape[0] = scan_shape[0] - 1
                 # Create the proper scan_positions without the flyback column
                 scan_positions = np.ravel_multi_index([ii.ravel() for ii in np.indices(scan_shape)],scan_shape)
-            
+                
             # Load any metadata
             metadata = {}
             if 'metadata' in f:
                 load_h5_to_dict(f['metadata'], metadata)
 
         scan_shape = scan_shape[::-1]
-
+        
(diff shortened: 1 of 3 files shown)


ercius:
This does not save a lot of time when loading data, but it saves ~2x the memory due to a long standing memory leak in `h5py` objects. Thus, for large data sets the option to load without the flyback can help with very large datasets and memory constraints.

Ready to be looked at @cjh1, @swelborn!

ercius:
Another idea is to use fancy indexing. I have not tested the speed yet:
```python
# Remove the flyback data
    to_keep = np.ones((scan_shape[0] * scan_shape[1]), dtype=bool)
    to_keep[scan_shape[0]::scan_shape[1]] = False
    data = frames[to_keep]
```

ercius:
There is a minimal effect on the speed of loading with either method. The more recent method would allow any "load map" to be used rather than just deleting the last (flyback) column. It seems like a useful feature.

python/stempy/io/sparse_array.py
```diff
@@ -188,19 +190,31 @@ def from_hdf5(cls, filepath, **init_kwargs):
...
             frame_shape = [frames.attrs[x] for x in ['Nx', 'Ny']]
-
-            scan_positions = scan_positions_group[()]
+            
+            if keep_flyback:
+                data = frames[()] # load the full data set
+                scan_positions = scan_positions_group[()]
+            else:
+                # Generate the original scan indices from the scan_shape
+                orig_indices = np.ravel_multi_index([ii.ravel() for ii in np.indices(scan_shape)],scan_shape)
```
swelborn, line 201: any reason to use `ii` for indices? instead of `i` for ex.

tests/test_sparse_array.py
```diff
@@ -710,6 +710,13 @@ def compare_with_sparse(full, sparse):
     assert np.array_equal(m_array[[False, True], 0][0], position_one)
 
 
+def test_keep_flyback(sparse_array_small):
```
swelborn, line 713: no test?

swelborn (commented):
I am not too familiar with sparse array class here, but my two cents in this review. perhaps @cjh1 can give a bit more feedback

python/stempy/io/sparse_array.py
ercius: I never use single letter variables.

I once spent a full day debugging something where `i` was already used for `sqrt(-1)`. Ever since, I always use `ii`, `jj`, `kk`, etc. for loops.

Also, when using the `pdb` [debugger](https://docs.python.org/3/library/pdb.html#pdbcommand-next), `c`, `q`, `n`, etc. are reserved for interacting with the debugger (continue, quit, next, etc.). That messed me up a few times as well.

tests/test_sparse_array.py
ercius: Looks like I started the test but did not finish it.

ercius: test added

ercius:
I think this is ready for review by someone else.

cjh1 (approved):
LGTM

cjh1:
@ercius I have cut a new release which contains this change. I have updated the images using by jupyterlab.

PR #318, opened 2024-08-22
Fix remove_flyback algorithm

ercius:
The current algorithm for removing the flyback doe snot properly account for multiple frames per probe position.

This new algorithm is designed to account for that problem.

Its not ready yet. I need to build new stempy version to test it.

### DIFF
--- a/python/stempy/io/sparse_array.py
+++ b/python/stempy/io/sparse_array.py
@@ -197,7 +197,7 @@ class SparseArray:
                 data = frames[()] # load the full data set
                 scan_positions = scan_positions_group[()]
             else:
-                num = frames.shape[0] // scan_positions_group.shape[0] # number of frames per probe position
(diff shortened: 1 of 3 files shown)


ercius:
This looks like it is working. Can we issue a release candidate with this and the other pull requests? I want to test it a little more thoroughly.

cjh1:
@ercius Sorry for the delayed response. As part of the CI for a PR a wheel is built ( just download the the artifacts ) you can use that for testing. The CI is currently failing because you are removing `scan_positions` and it is still referenced in the version 3 block.

ercius:
@cjh1 Thanks for finding my error. I've added in scan_positions and will grab the artifact to more fully test this new algorithm.

ercius:
@cjh1 I think this is good to merge. Can you take a look?

cjh1 (approved):
LGTM

PR #328, opened 2024-10-03
Docstring update

ercius:
Update the docstring to the electron counting code. Its best to use a simple np.ndarray.


swelborn (approved):
lgtm

PR #330, opened 2024-12-19
aarch64 build wheel

swelborn:


cjh1 (approved):

cjh1:
@swelborn Wow, that took a while to build!

swelborn:
Indeed. I think we will have access to arm runners soon? https://github.com/github/roadmap/issues/970 says Q1 2025.

PR #332, opened 2025-05-30
Fix docker CI builds

swelborn:
`mpi4py` defaults to using setuptools as backend. Changing this to `scikit-build-core` was the key here. 

Other things are useful anyway, so I keep them here - upgrading action version, and not using cache directory (smaller image sizes).


cjh1 (approved):
LGTM

PR #341, opened 2025-08-27
Set threshold directly in electron_count

ercius:
I think it might be more convenient to set the threshold directly when electron counting sometimes. Here I add two new keywords and some logic to test for those keywords. 

I have not tested this yet, but the code looks ok to me.

@cjh1 @swelborn What do you think of this method for checking the keywords? Also, do we have a test for electron_counting? We would need a raw data set available to the tests which is a large amount of data. 

Ill test this once I can get the docker container built with this code.


ercius:
Those first three commits are from my old weird history. Ill resend this PR when I can clean up my history.

PR #348, opened 2025-09-05
use py311 for docker builds

swelborn:


swelborn:
i see @ercius already doing this on #346

PR #350, opened 2025-12-09
Improve com crop

ercius:
I previously used a square mask to reject electrons outside the central beam. This is not ideal. I thought that making a radial mask would be slow, but it turns out to be slightly faster per frame. I estimate about 4 seconds saved for a medium sized 1kx1k sparse dataset and a more accurate measurement. Win-win!

This needs to be tested before merging. Ill update when it looks good to merge.

### DIFF
--- a/python/stempy/image/__init__.py
+++ b/python/stempy/image/__init__.py
@@ -552,8 +552,8 @@ def _com_sparse_v0(array, crop_to=None, init_center=None, replace_nans=True):
             
             if crop_to is not None:
                 # Crop around the initial center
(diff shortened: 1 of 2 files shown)


ercius:
@cjh1 This looks good to go on my side. I tested it on some other data sets. I am not sure how to run the tests though. 

If this is good then we can merge and cut a new release.

cjh1 (approved):
LGTM

cjh1:
@ercius Once this is merged I will cut a release.

cjh1:
I have cut a release that contains these changes.

PR #351, opened 2026-01-08
Update wheel building workflow

ercius:
We are using python 3.10 to build and getting warnings like this:
"cibuildwheel 3 will require Python 3.11+, please upgrade the Python version used to run cibuildwheel. This does not affect the versions you can target when building wheels."
This uses python 3.12 to build.

Also, update build workflow to support CPython >= 3.10 by skipping older EOL versions of python. This should speed up the build process and reduce wasted resources.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -9,8 +9,8 @@ on:
   pull_request:
 
 env:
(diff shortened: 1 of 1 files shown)


ercius:
Ill close this and work from main.

Issue #353, opened 2026-02-13
get_scan_path bug


ercius:
There is a problem with `get_scan_path` when standard and centered data are in the same directory and only a `scan_id` is provided

[This line](https://github.com/OpenChemistry/stempy/blob/2a08eeb4f22b417d05420b7618620e9e65bd76e6/python/stempy/contrib/__init__.py#L66) needs to be updated to match proceeding lines:
`file_pattern = f"*_id{scan_id}*_electrons{file_suffix.value}.{filetype}"`

PR #356, opened 2026-08-28
account for when centered and normal data are in the same directory a…

ercius:
…nd only scan_id is given


ercius:
This fixes #353

ercius:
still needs testing. I need it to build first.

ercius:
This only affects the older `data_scan_`... files. 

Also, its tested and works.

@cjh1 Take a look and merge when its good.

cjh1 (approved):
LGTM

